PR #301, opened 2022-03-02
Translate statistics page

seppeljordan:
I translated the "Statistiken" page. It was necessary to introduce `ngettext` to the translator interface. I chose to move some of the translation logic to the presenter since this is where the translation should happen anyway.


seppeljordan:
Paid by myself

PR #302, opened 2022-03-02
Avoid running unittests in CI twice

seppeljordan:
Before this change the CI pipeline steps using the `nix` build tool were running the unittests twice: Once as part of `nix build` and then again as part of `./run-checks`. To avoid this I introduced an optional argument to the `./run-checks` program that disables unittests.


mi01 (approved):

seppeljordan:
I just paid the PR.

PR #307, opened 2022-03-02
Cleanup deployment files

seppeljordan:
This PR moves the existing settings files into the `arbeitszeit_flask` package. Additionally all modules in `arbeitszeit_flask`, `arbeitszeit_web` and `arbeitszeit` are now packaged via setuptools. This means that one could now install the app via `pip install`.

This change was done in preparation for building docker images.

At least in theory the production instance of the app at heroku should "just work" without any modifications to the deployment settings.

Developers need to change their environment to
```
export ARBEITSZEIT_APP_CONFIGURATION="$PWD/arbeitszeit_flask/development_settings.py"
```
This is documented in `README.rst` already.


sloschert (approved):
PR has been paid.

PR #308, opened 2022-03-03
Translate profile page

seppeljordan:
This PR makes the profile pages for members and companies translatable.

PlanID: 1906a785-a23b-44b3-a599-96ee43dd388c


seppeljordan:
Paid this myself.

PR #322, opened 2022-03-12
Prevent negative work certificate transfer

### LINKED ISSUE #320: Companies can transfer negative work certificates to workers
Only a **positive** value should be possible.

sloschert:
fixes #320

This PR makes it impossible to transfer negative work certificates to workers.

Along the way I had to refactor the corresponding use case quite extensively. That is, I created a controller and presenter and moved the route into it's separate view. 

The presenter has two different methods, one to render warnings from the controller and one to render the use case response.


seppeljordan (approved):
Nice work. Certificates have been transferred.

PR #330, opened 2022-03-17
Prevent-negative-plan-costs

### LINKED ISSUE #321: Companies can specify negative p, r, a in plan creation process
Only **positive** values should be allowed.

sloschert:
fixes #321

This PR prevents that companies specify negative plan costs. It does this 1) in the html ``input`` element and 2) by assertions in the controller.

I refrained from refactoring the controller any further because I plan to propose some changes related to plan creation backend soon.

### DIFF
--- a/arbeitszeit/use_cases/create_plan_draft.py
+++ b/arbeitszeit/use_cases/create_plan_draft.py
@@ -32,6 +32,11 @@ class CreatePlanDraft:
     datetime_service: DatetimeService
 
     def __call__(self, request: CreatePlanDraftRequest) -> CreatePlanDraftResponse:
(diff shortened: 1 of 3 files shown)


seppeljordan:
I think that the check should go into the use case, since this is a business logic concern, and not just mere user input validation.

seppeljordan (approved):

sloschert:
> Anyway, this could be merged for now. But we should not forget to implement the proper checks in the use case. If you want to go ahead and merge this as is you could create an issue so we don't forget.

Allright, as I said, I have assert statements in the use case. But I will create an issue to implement proper error handling, soon.

PR #331, opened 2022-03-18
Simplify plan renewal

sloschert:
I think that the "create plan" procedure is unnecessarily complicated - especially the "plan renewal" part of it. "Plan renewal" should simply mean to take the plan info of an expired plan and copy it into a new plan draft. 

But right now the old plan is marked as "renewed" as well. This is unnecessarily complex: In my opinion, neither a future plan approval workflow nor accumulation control workflow will need the category of a "renewed plan". *) 

That's why I propose in this PR to delete the "Plan.renewed" attribute. "Renewing" a plan still is possible in the simpler meaning of copying the plan content to a new draft. Do you agree?

In an upcoming PR i would then refactor some of the parts of the "create plan" procedure. 

------
*) The category of a "renewed plan" was originally implemented because a difference between renewing an expired plan (potentially without any changes to the plan attributes) and filing a completely new plan was desired. 
The idea was that renewing a plan without any changes would allow for automatic plan approval - but this is not true: criteria for plan approval should be, if plans have been met met, not if changes have been made to the plan.
Accumulation control originally was another argument in favor of having this category ("how much have increased the company's use of means of production?") - but accumulation control can also be done with completely new plans.

### DIFF
--- a/arbeitszeit/entities.py
+++ b/arbeitszeit/entities.py
@@ -139,7 +139,6 @@ class Plan:
     approval_reason: Optional[str]
     is_active: bool
     expired: bool
-    renewed: bool
     activation_date: Optional[datetime]
     expiration_relative: Optional[int]
     expiration_date: Optional[datetime]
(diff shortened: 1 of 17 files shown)


seppeljordan (approved):
Great change. Certificates were transferred.

Issue #312, opened 2022-03-03
Make all user facing strings translatable


seppeljordan:
This issue tracks the translation effort. We need to check every html template and every presenter for untranslated strings.

- [ ] arbeitszeit_flask/templates/base.html
- [ ] arbeitszeit_flask/templates/member/query_companies.html
- [ ] arbeitszeit_flask/templates/member/show_company_work_invite_details.html
- [ ] arbeitszeit_flask/templates/member/query_plans.html
- [ ] arbeitszeit_flask/templates/member/read_message.html
- [ ] arbeitszeit_flask/templates/member/plan_summary.html
- [ ] arbeitszeit_flask/templates/member/company_summary.html
- [ ] arbeitszeit_flask/templates/member/list_messages.html
- [ ] arbeitszeit_flask/templates/member/my_account.html
- [ ] arbeitszeit_flask/templates/member/my_purchases.html
- [ ] arbeitszeit_flask/templates/member/pay_consumer_product.html
- [ ] arbeitszeit_flask/templates/member/help.html
- [ ] arbeitszeit_flask/templates/member/404.html
- [ ] arbeitszeit_flask/templates/member/coop_summary.html
- [ ] arbeitszeit_flask/templates/member/statistics.html
- [ ] arbeitszeit_flask/templates/member/profile.html
- [ ] arbeitszeit_flask/templates/base_company.html
- [ ] arbeitszeit_flask/templates/macros/read_message.html
- [ ] arbeitszeit_flask/templates/macros/company_summary.html
- [ ] arbeitszeit_flask/templates/macros/list_messages.html
- [ ] arbeitszeit_flask/templates/macros/show_notifications.html
- [ ] arbeitszeit_flask/templates/macros/coop_summary.html
- [ ] arbeitszeit_flask/templates/company/query_companies.html
- [ ] arbeitszeit_flask/templates/company/list_all_transactions.html
- [ ] arbeitszeit_flask/templates/company/account_p.html
- [ ] arbeitszeit_flask/templates/company/query_plans.html
- [ ] arbeitszeit_flask/templates/company/read_message.html
- [ ] arbeitszeit_flask/templates/company/plan_summary.html
- [ ] arbeitszeit_flask/templates/company/my_accounts.html
- [ ] arbeitszeit_flask/templates/company/company_summary.html
- [ ] arbeitszeit_flask/templates/company/list_messages.html
- [ ] arbeitszeit_flask/templates/company/transfer_to_company.html
[...]

PR #339, opened 2022-03-28
Make more strings translatable

sloschert:
Making six more files from list in #312 translatable.

Plan-ID: e8e5c56f-9e2d-497d-a896-a26bc41e1c16

### DIFF
--- a/arbeitszeit_flask/templates/company/my_purchases.html
+++ b/arbeitszeit_flask/templates/company/my_purchases.html
@@ -9,7 +9,7 @@
   <div class="box has-background-info-light has-text-info-dark has-text-centered">
     <div class="icon"><i class="fas fa-info-circle"></i></div>
     <p>
-      {{ gettext("Here you can review your past purchases.") }}<br>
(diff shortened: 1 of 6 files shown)


arbeitszeit_flask/templates/company/my_purchases.html
```diff
@@ -4,12 +4,12 @@
...
 <div class="content">
   <h1 class="title">
-    Meine vergangenen Käufe:
+    {{ gettext("My past purchases") }}
   </h1>
   <div class="box has-background-info-light has-text-info-dark has-text-centered">
     <div class="icon"><i class="fas fa-info-circle"></i></div>
     <p>
-      Hier siehst du deine vergangenen Käufe.<br>
+      {{ gettext("Here you can review your past purchases.") }}<br>
```
knit-bee, line 12: Maybe use 'view' instead of 'review'

knit-bee (approved):
I would change one string in the my_purchases.html file, but I think it is also okay as it is. The rest looks good to me.

sloschert:
Thanks for the review, I paid the PR.

PR #342, opened 2022-03-29
delete work.html

sloschert:
`work.html` is not needed anymore. Deleting it.

No payment needed.


seppeljordan (approved):
Thanks for the cleaning up

PR #343, opened 2022-04-01
Refactor draft creation

### LINKED ISSUE #341: Proper checks to prevent negative values in `CreatePlanDraft`
Proper checks are missing to prevent negative p, r, a, amount and timeframe values as inputs to the use case `CreatePlanDraft` (`arbeitszeit/use_cases/create_plan_draft.py`). 

See discussion in #330.

sloschert:
fixes #341

This PR proposes some changes around the draft creation process. 

Firstly, it implements proper checks in the `CreatePlanDraft` use case as described in the issue mentioned above.

Then  it simplifies the draft creation process in the flask module. It merges two routes into one and moves it into a view class.  It implements tests for the new view class. 

There could be done some more improvements in the future (e.g. separating concerns, adding controller and presenter), but i don't want to bloat this PR.


seppeljordan (approved):
Great change.

seppeljordan:
Certificates transferred.

PR #344, opened 2022-04-05
Plot for prd account

sloschert:
This PR adds a plot to the prd account overview. The plot visualizes the account movements. 

To achieve this, this PR introduces the `arbeitszeit_web.plotter.py` module and the corresponding implementation on flask-level, `arbeitszeit_flask.flask_plotter.py`. 

Instead of using matplotlib directly, the presenter calls `FlaskPlotter().create_line_plot(x, y, size)`. This has the advantage that we are not bound to matplotlib in the future but can more easily change the plotting backend when we need to. 

This plotter might be used similarily for the other account overviews as well as for the statistics page.

The FlaskPlotter itself is not tested, but it's inputs are. 

Plan-ID: e8e5c56f-9e2d-497d-a896-a26bc41e1c16

image


arbeitszeit/use_cases/show_prd_account_details.py
```diff
@@ -19,10 +19,17 @@ class TransactionInfo:
     purpose: str
 
 
+@dataclass
+class PlotDetails:
+    dates: List[datetime]
```
seppeljordan, line 24: I think that the field name `dates` is missleading since the user would expect a `date` object. how about `timestamps`?

arbeitszeit_flask/flask_plotter.py
```diff
@@ -0,0 +1,25 @@
...
+from typing import List, Tuple
+
+from matplotlib.figure import Figure
+
+
+class FlaskPlotter:
+    def create_line_plot(
+        self, x: List[datetime], y: List[Decimal], size: Tuple[int, int] = (10, 5)
+    ) -> str:
+        fig = Figure()
```
seppeljordan, line 14: I think that the shortened variable names are making it harder to understand what is going on here.

seppeljordan (changes requested):
I have no major complaints, quite the opposite: I think that this is a well done change. BUT I think that some stylistic things should be improved to help future programmers in understanding the code as it is right now.

arbeitszeit_flask/flask_plotter.py
sloschert: Changed `size` to `fig_size` - is that better? Regarding x and y: Not quite sure which name would make it more comprehensive. In plotting x and y are common variable names. Would you prefer `x_values` or `x_axes_values`?

seppeljordan (approved):

seppeljordan:
Thanks for the improvement. Certs have been transferred.

PR #346, opened 2022-04-08
Introduce dependency injection to presenter tests

seppeljordan:
For the sake of unification I propose to use dependency injection for conducting the presenter tests. IMO this will increase modularity and reduce friction when further developing the application. No functional changes were implemented, neither to prod code nor to the tests.

Plan: 1906a785-a23b-44b3-a599-96ee43dd388c


sloschert:
I like it. Certificates transferred.

sloschert (approved):

PR #347, opened 2022-04-08
Docker image

seppeljordan:
This PR implements the build "recipe" for a docker container image via the nix build tool. The docker image is functional only in combination with a https reverse proxy (like nginx for example) since the production settings are used. These production settings disallow access via http and only allow https.

Plan: 1906a785-a23b-44b3-a599-96ee43dd388c


sloschert (approved):
Certificates transferred.

PR #348, opened 2022-04-09
Changes to translation procedure

sloschert:
This PR proposes some changes to the translation procedure:

- It deactivates babel's fuzzy matching (babel was trying to find translations of similar strings in our catalog). That was not working well.
- All flask-related translation files are now in the `arbeitszeit_flask` module, not on top level of the application
- There are now simple flask custom commands to execute the rather lenghty babel commands
- README has been adapted

Plan: e8e5c56f-9e2d-497d-a896-a26bc41e1c16

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -114,26 +114,46 @@ In development mode you can run it manually in the CLI.
 Translation
 ===========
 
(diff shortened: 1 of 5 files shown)


seppeljordan (approved):
👍🏼 Certificates have been transferred.

PR #350, opened 2022-04-09
Making strings translatable

sloschert:
Making user facing strings in four more presenters translatable.


seppeljordan (approved):
Certificates transferred.

PR #351, opened 2022-04-10
Refactor some more presenters to use dependency injection

seppeljordan:
This is pure refactoring. No features nor testing logic was changed.

I changed a bunch of presenter tests to use dependency injection instead of direct construction.


sloschert (approved):
Certs transferred.

PR #352, opened 2022-04-10
Update (nix) dependencies

seppeljordan:
This PR updates the nixpkgs revision that is used and also the `requirements.txt` file.


sloschert (approved):
Certs transferred.

PR #353, opened 2022-04-11
Refactoring of html templates

sloschert:
This PR makes some long due changes to html files and the jinja template structure, sorry for the huge and convoluted PR. I recognize that it might not be possible to review this in detail. Much of it is pure formatting.
 
1) Changes to jinja structure: The main jinja template is called `base.html`. There are now three child templates inheriting from the base: `base_member.html`, `base_company.html` and `base_general_pages.html`. The last template is for general purpose-pages that are not associated to the member or company-side of the app.  
2) All three child templates are calling the `show_notifications()`-macro function (after rendering of the navigation bar). As a consequence, we do not need to call this function in the individual templates anymore, which prevents code duplication. 
3) The notification box is shown after the navbar. It now has a "close"-Button that works with Javascript. 
4) Some clean-up: 

- The templates of the "auth"-module have been moved in the corresponding `templates`-folder.
- Moved macros into the `macros`-folder.

---- 

What could be done in the future:

- Standardize some existing notification implementation that differ from our standard implementation. E.g. in member/pay_consumer_product

----

Plan: e8e5c56f-9e2d-497d-a896-a26bc41e1c16


seppeljordan (approved):
Great change :)
Certs have been transferred.

PR #355, opened 2022-04-19
Add account balances to company summary

### LINKED ISSUE #354: Metrics for plan control
We need some usefull metrics that allow "plan control". Here are some ideas for possible metrics. 

---

The main question for plan control is: "How good did a company fulfill its plans in the past?" 

"Fulfillment" means that the company in fact used as much p, r and a as specified in its plans and gave away all the product (=p+r+a). Ideally neither more nor less.

Possible mectrics can be of two kind:  
- metrics on plan-level, which assess how good a company did fulfill a single plan (or all plans aggregated)
- mectrics on company-level (less precise)
 
 
1.

Plan-specific metrics answer the question how good a company did fulfill a specific plan. Unfortunately, most data we have is not plan-specific: Companies "pay" p and r and transfer a to workers without specifying for which plan these means of production or work was used. The reason for this is that it would be overwhelming for companies to be obliged to pinpoint which p, r and a transactions are related to which plan exactly. Hence, we don't ask for this information.  

In consequence, we only have one plan-specific metrics: The difference between expected "sales" and real "sales" (per plan). We might call this metrics "sales deviation". 

To make this absolute metrics comparable across different companies and plans, we can create a relative metrics: "sales deviation relative to expected sales"

2.

On the other hand we have data on company-level. For example: In the last five years, has the company spent more on p than specified in all its plans, less or exactly the amount specified? This information on company-level does not tell us anything about specific plans, but it gives us an average number over a longer timeperiod or several plans. 

These overall metrics are calculated by subtracting the real overall value from the expected overall value. Examples:
- p: A company has planned overall p=100. It only "paid" means of production of 50. -> The "means of production deviation" = +50 hours
- r: A company has planned overall r=100. It "spent" actually 200 in raw materials. -> The "raw materials deviation" = -100 hours
- a: A company has planned overall a=100. The FIK is 0.8. It is expected to send 80 certificates to workers. It only sent 40 certificates. -> The "certificate deviation" = +40 hours
- prd: A company has planned overall to give away p+r+a=300 hours of product. It gave away only 150 hours of product. -> The "sales deviation" = -150 hours

We already have these overall metrics in our app: They are used in the account balances of the accounts p, r, a and prd on the page "my account".

To make these overall metrics comparable across different companies we can create relative metrics:

- p: "means of production deviation in relation to planned p" (= in relation to received credit for p)
- r: "raw materials deviation in relation to planned r" (= in relation to received credit for r)
- a: "certificate deviation in relation to received certificates" (= in relation to planned a * FIK)
- prd: "sales deviation in relation to expected sales" (= in relation to planned p+r+a)

sloschert:
Fixes #354

This PR posposes an implemention of the plan-specific and company-wide metrics described in issue #354.

The metrics for all companies are accessible for all members and companies via the `company_summary`-page.

image


seppeljordan (approved):
Great change. Certs transferred.

PR #356, opened 2022-04-20
New design and metrics for global statistics

sloschert:
This PR implements a new design for the statistics page, using bulmas [tile](https://bulma.io/documentation/layout/tiles/) grid system instead of simple tables. Inspiration came from data visualization tools like Metabase. 

Some new global metrics on the statistics page make use of a new Plotter method, `create_bar_plot()` and are visualized as bar plots. 
 
The PR introduces also a new module, `arbeitszeit_web.colors.py`. It's implementation in flask, `arbeitszeit_flask.flask_colors.py`, allows to store default colors for our webapp.

The current set up is resource-intensive and not very scalable because it creates global data for every request from scratch. We should discuss alternatives.   

image


arbeitszeit/use_cases/get_statistics.py
```diff
@@ -24,15 +34,48 @@ class GetStatistics:
...
         )
+
+    def _count_certificates_and_available_product(self) -> Tuple[Decimal, Decimal]:
+        """
+        available product is sum of prd account balances *(-1)
+        available certificates is sum of company work account balances and sum of member account balances
+        """
+        certs_in_company_accounts = Decimal(0)
+        available_product = Decimal(0)
+        all_companies = self.company_repository.get_all_companies()
```
seppeljordan, line 66: This code would be a little bit cleaner if two methods were factored out: `_count_certs_and_products_from_companies() -> Tuple[int,int]` and `_count_certs_in_member_accounts() -> int`

tests/flask_integration/test_member_repository.py
```diff
@@ -68,3 +68,15 @@ def test_get_by_id_returns_none_when_member_does_not_exist(
 ) -> None:
     member = repository.get_by_id(uuid4())
     assert member is None
+
+
+@injection_test
+def test_that_all_members_can_be_retrieved(
```
seppeljordan, line 74: I think we could use an additional test that checks if the number of returned members is equal to the number of created members.

arbeitszeit_web/get_statistics.py
```diff
@@ -1,36 +1,48 @@
...
 
+    barplot_certificates: str
+    barplot_means_of_production: str
+    barplot_plans: str
+
 
 @dataclass
 class GetStatisticsPresenter:
-    translator: Translator
+    trans: Translator
```
seppeljordan, line 31: Why did you shorten the name `translator`? When reading this I think less work on the part of the reader is required with the non-shortened version.

arbeitszeit_web/colors.py
```diff
@@ -0,0 +1,26 @@
...
+
+    @property
+    def danger(self) -> str:
+        ...
+
+    @property
+    def success(self) -> str:
+        ...
+
+    def get_dict(self) -> Dict[str, str]:
```
seppeljordan, line 25: I think that this method deserves a better name. I read `Colors.get_dict` and have no idea what this method is supposed to do.

tests/plotter.py
```diff
@@ -1,10 +1,20 @@
...
+
+    def create_bar_plot(
+        self,
+        x_coordinates: List[Union[int, str]],
+        height_of_bars: List[Decimal],
+        colors_of_bars: List[str],
+        fig_size: Tuple[int, int],
+        y_label: Optional[str] = None,
+    ) -> str:
+        ...
```
seppeljordan, line 20: The return value of the method should be a string. As it is right now `None` is returned.

arbeitszeit_flask/templates/company/account_prd.html
```diff
@@ -45,6 +45,7 @@ <h1 class="title">
                 </tr>
                 {% endfor %}
                 {% endif %}
+
```
seppeljordan, line 48: This empty line should be removed

arbeitszeit_flask/templates/macros/statistics.html
```diff
@@ -1,114 +1,86 @@
...
-                </div>
+        </div>
+        <div class="tile is-parent is-4 has-text-centered">
+            <div class="tile is-child box">
+                <h1 class="title">{{ view_model.average_timeframe_days }}</h1>
+                <div class="subtitle">{{ gettext("Average plan duration") }}</div>
             </div>
         </div>
     </div>
+
```
seppeljordan, line 83: This empty line should be removed.

seppeljordan (changes requested):
Overall a massive win for the working class. But I had some stylistic complaints.

arbeitszeit_web/get_statistics.py
```diff
@@ -41,7 +53,60 @@ def present(self, use_case_response: StatisticsResponse) -> GetStatisticsViewMod
...
+                    use_case_response.certificates_count,
+                    use_case_response.available_product,
+                ],
+                colors_of_bars=[self.colors.primary, self.colors.info],
+                fig_size=(5, 4),
+                y_label=self.translator.gettext("Hours"),
+            ),
+            barplot_means_of_production=self.plotter.create_bar_plot(
+                x_coordinates=[
+                    self.translator.gettext("Feste PM"),
```
seppeljordan, line 79: I did not catch this on my last review. But you mixed german translation ids with english translation ids.

seppeljordan (approved):
Thanks for the change. Certs transferred

PR #357, opened 2022-04-20
Refactor presenter tests to utilize dependency injection

seppeljordan:
This is a continuation of the ongoing refactoring effort. Mainly some more presenter tests were refactored to use dependency injection in their object construction. The long term goal here is to reduce coupling between the test cases and the object construction for classes that are not under test.


sloschert (approved):
Thanks, certs transferred!

PR #361, opened 2022-05-01
Seppeljordan/refactor registration

seppeljordan:
I reworked the sending of registration emails for better testing and more extensive testing.

This PR conceptualizes the sending of emails as just another process of displaying data to a user. Hence the introduction of the `MemberRegistrationMessagePresenter` and `CompanyRegistrationMessagePresenter`. These presenters are called directly by the respective use cases. This deviates somewhat from our usual approach. I think that this deviation is actually an improvement since it makes the whole architecture more flexible with regards to different "data outlets".

A consequence of this change is that the refactored use cases define the "call conventions" for the presenters as a `Protocol`. This may seem like a contradiction to our current architecture since it looks like there is presenter code inside of the `arbeitszeit/` folder. I'd argue though that these protocols are just an iteration on the concept of use case response models.

Further refactoring needs to be done since the registration logic is still somewhat messy.


arbeitszeit_web/presenters/registration_email_presenter.py
```diff
@@ -0,0 +1,41 @@
...
+    def render_to_html(self, confirmation_url: str) -> str:
+        ...
+
+
+@dataclass
+class RegistrationEmailPresenter:
+    email_sender: MailService
+    address_book: UserAddressBook
+    member_email_template: RegistrationEmailTemplate
+    company_email_template: RegistrationEmailTemplate
```
sloschert, line 20: `company_email_template` is not used - in case you overlooked it. But probably we could leave it as is because later we surely will have different templates for companies and members.

tests/email.py
```diff
@@ -0,0 +1,48 @@
...
+            Email(
+                subject=subject,
+                recipients=recipients,
+                html=html,
+                sender=sender,
+            )
+        )
+
+
+@dataclass
```
sloschert, line 35: does this class have to be a `dataclass`?

seppeljordan: it was unnecessary. I removed the decorator.

arbeitszeit_web/presenters/registration_email_presenter.py
seppeljordan: Thanks for catching this.

sloschert (approved):
Thanks, great work. The structural changes seem like improvements to me as well. Certs transferred.

PR #364, opened 2022-05-03
Refactor /member/signup

seppeljordan:
This PR refactors the route `/member/signup`.

# `RegisterMemberPresenter` was introduced

This presenter is a little different from other presenters that we already implement. Its responsibility is not only to render html to the user but also to sign the user in when the registration attempt was successful. I extended the concept of a `Session` to accommodate the change. Furthermore the error message that appears when an email is already registered is now translatable.

# `SignupMemberView` was implemented

This view is functionally identical to the code that was previously in the route. The newly implemented presenter replaced the code that handled successful form submissions.

I'm looking forward to feedback on this change since I would like to adopt this approach for the route `/company/register` and later `/accountant/register`.

### DIFF
--- a/arbeitszeit_web/presenters/register_member_presenter.py
+++ b/arbeitszeit_web/presenters/register_member_presenter.py
@@ -1,0 +1,34 @@
+from dataclasses import dataclass
+
+from arbeitszeit.use_cases.register_member import RegisterMemberUseCase
+from arbeitszeit_web.register_member import RegisterForm
+from arbeitszeit_web.session import Session
+from arbeitszeit_web.translator import Translator
+
+
+@dataclass
+class RegisterMemberViewModel:
+    is_success_view: bool
+
+
+@dataclass
+class RegisterMemberPresenter:
+    session: Session
+    translator: Translator
+
+    def present_member_registration(
+        self, response: RegisterMemberUseCase.Response, form: RegisterForm
+    ) -> RegisterMemberViewModel:
+        if response.is_rejected:
+            if (
+                response.rejection_reason
+                == RegisterMemberUseCase.Response.RejectionReason.member_already_exists
+            ):
+                form.add_email_error(
+                    self.translator.gettext("This email address is already registered.")
+                )
+            return RegisterMemberViewModel(is_success_view=False)
+        else:
+            email = form.get_email_string()
+            self.session.login_member(email)
+            return RegisterMemberViewModel(is_success_view=True)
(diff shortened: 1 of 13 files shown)


arbeitszeit_web/presenters/register_member_presenter.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def present_member_registration(
+        self, response: RegisterMemberUseCase.Response, form: RegisterForm
+    ) -> RegisterMemberViewModel:
+        if response.is_rejected:
+            if (
+                response.rejection_reason
+                == RegisterMemberUseCase.Response.RejectionReason.member_already_exists
+            ):
+                form.add_email_error(
```
sloschert, line 27: wouldn't it be more consistent if we just use our notification system (`notifier.display_warning()`) in this case instead of adding an error and iterating over the lists of errors in the template? of course we would lose the relationship with a specific input field, but the registration field has only 3 inputs.

arbeitszeit_web/presenters/register_member_presenter.py
```diff
@@ -0,0 +1,34 @@
...
+                    self.translator.gettext("This email address is already registered.")
+                )
+            return RegisterMemberViewModel(is_success_view=False)
+        else:
+            email = form.get_email_string()
+            self.session.login_member(email)
```
sloschert, line 33: i guess conceptually a better place for the log-in would be in the use case, right? is it a pragmatic decision to deviate from the rule here?

sloschert:
I just had two questions, but it's fine for me how it is.

arbeitszeit_web/presenters/register_member_presenter.py
seppeljordan: I'd argue that that "logging the user in" is something that should be handled in the `arbeitszeit_web` package since it is specific to the web platform. If the application was a desktop program then the "logging in" wouldn't make sense necessarily.

arbeitszeit_web/presenters/register_member_presenter.py
seppeljordan: We could move the notification to the `Notifier`. We should keep in mind that we might want to do validation on the other fields as well, e.g. password strength validation. This means that we could "regret" making this message a global notification later down the line.

arbeitszeit_web/presenters/register_member_presenter.py
sloschert: Ok!

sloschert (approved):

sloschert:
Certs transferred.

PR #365, opened 2022-05-04
Let FlaskSession handle login and logout

seppeljordan:
This PR refactors the auth routes with respect to logging in and logging out.  All login related functions are now handled by `FlaskSession`. Unfortunately I was not able to remove some references to `session["user_type"]` from the rest of the code (which would be desirable IMO).

### DIFF
--- a/arbeitszeit_flask/flask_session.py
+++ b/arbeitszeit_flask/flask_session.py
@@ -3,14 +3,21 @@ from typing import Optional
 from uuid import UUID
 
 from flask import session
-from flask_login import current_user, login_user
+from flask_login import current_user, login_user, logout_user
 
-from arbeitszeit_flask.database.repositories import MemberRepository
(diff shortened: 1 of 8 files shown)


arbeitszeit_flask/flask_session.py
```diff
@@ -3,22 +3,38 @@
...
         try:
             return UUID(current_user.id)
         except AttributeError:
             return None
 
-    def login_member(self, email: str) -> None:
+    def login_member(self, email: str, remember: bool = False) -> None:
         member = self.member_repository.get_member_orm_by_mail(email)
-        session["user_type"] = "member"
         login_user(member)
```
sloschert, line 30: I think you forgot to pass on the `remember` argument.

sloschert (approved):
I had one remark, apart from that it's approved. Certs transferred.

seppeljordan:
Thanks for catching the error.

PR #366, opened 2022-05-05
Refactor /company/signup

seppeljordan:
I refactored the route /company/signup similarly to the route /member/signup,


sloschert (approved):
Certs transferred.

PR #367, opened 2022-05-06
Use flask endpoints for plot creation

### LINKED ISSUE #359: Plot does not render on production server
The plot of prd account does not render on the production server when heading to `company/my_accounts/account_prd`. It only shows the alt text.

sloschert:
fixes #359 

This PR proposes a solution to the security issue described in #359. Presenters do not insert a base64 string into the templates anymore but return an url to an endpoint where plots are created. This makes the plot rendering more complicated. 

I'm interested in your opinion if there is something I could improve. But from what I read and from what we discussed so far this seems to be more or less the way to go. 

Some remarks:

- There is a new flask blueprint called "plots".
- For test performance reasons I wrote integration tests of url indexes only for companies, not for members (in `tests/flask_integration/test_url_index.py`).
- In the current setting it is possible for members as well as for other companies to access a company's PRD account plot (by manually typing the url). The only access restriction is that users must be logged in. I think that is fine, as this should be probably public information.
- In route "line_plot_of_company_prd_account" there is a second call to Use Case `ShowPRDAccountDetails` (a first call takes place in the presenter). If we want to prevent this we could split the Use Case in two parts: One for showing the transactions plus the current balance, and a second one for just showing the prd account plot.

### DIFF
--- a/arbeitszeit_flask/url_index.py
+++ b/arbeitszeit_flask/url_index.py
@@ -1,3 +1,4 @@
+from decimal import Decimal
 from uuid import UUID
 
 from flask import url_for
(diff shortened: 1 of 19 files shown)


arbeitszeit_flask/url_index.py
```diff
@@ -42,6 +43,40 @@ def get_confirmation_url(self, token: str) -> str:
             endpoint="auth.confirm_email_member", token=token, _external=True
         )
 
+    def get_global_barplot_for_certificates_url(
```
seppeljordan, line 46: I looks to me like the implementations for the MemberUrlIndex and the CompanyUrlIndex are identical. I would suggest to implement one single PlotUrlIndex to reduce code duplication.

seppeljordan (changes requested):
I think the `*UrlIndex` implementations should be unified into one single implementation since there is no member/company specific logic in them.

The rest of this PR looks great.

sloschert:
Plan ID:  e8e5c56f-9e2d-497d-a896-a26bc41e1c16

seppeljordan (approved):
Great work. Certs have been transferred.

PR #369, opened 2022-05-08
Include templates in package data

seppeljordan:
This PR is about packaging. With the change proposed by this PR the templates used by `arbeitszeitapp` are also included in source and binary distributions of the app.


sloschert (approved):
Certs transferred.

PR #373, opened 2022-05-12
Fix-pytest-warnings

### LINKED ISSUE #309: Remove instances of using ngettext with Decimal
Currently the test suite raises warnings about the usage of ngettext in combination with Decimal. We should remove those instances and replace them with a simple call to `Translator.gettext`. This means that we loose singular forms in cases where `n = Decimal('1.0')`. Since this case almost never would occur this should be okay.

sloschert:
fixes #309

This PR fixes two different pytest warnings: 

- warnings about the usage of ngettext in combination with Decimal
- a warning that was raised because of the naming of `TestColors`

Plan ID: 26f2e3f6-233a-4ef1-a1a0-39c5d411b38c


seppeljordan (approved):
Thanks. Certs transferred.

PR #375, opened 2022-05-13
render transfer_to_company w/ empty form when GET

### LINKED ISSUE #325: Reset form when successfully paying means of production
When a company pays means of production ("Konten und Käufe" > "Bezahlen Produktionsmittel") then the form does not reset after successfully submitting the form. I should reset to a blank state though to prevent accidentally submitting the form twice.

### LINKED ISSUE #326: Certificates are transferred multiple times when when user uses "Back" navigation in browser
What was observed:

I transferred some certificates via the "Bezahlen Produktionsmittel" form. After doing so successfully I clicked the "Back" button in my browser (firefox). Afterwards I checked my accounts and noticed that the last payment had been made 3 times in one minute in addition to the original (intended) payment.

What I expected:

I expected the browser navigation to not resubmit the payment form to the server.

sloschert:
fixes #325
fixes #326

This PR should fix both form issues when paying means of production. 

@seppeljordan please confirm that issue #326 is fixed for you as well.

Plan ID: 26f2e3f6-233a-4ef1-a1a0-39c5d411b38c
(please add this plan to our cooperation as well)

### DIFF
--- a/arbeitszeit_flask/views/pay_means_of_production.py
+++ b/arbeitszeit_flask/views/pay_means_of_production.py
@@ -18,20 +18,20 @@ class PayMeansOfProductionView:
     presenter: PayMeansOfProductionPresenter
     template_renderer: UserTemplateRenderer
 
-    def respond_to_get(self):
-        return Response(self._render_template(), status=200)
+    def respond_to_get(self, form: PayMeansOfProductionForm):
+        return Response(self._render_template(form), status=200)
 
     def respond_to_post(self, form: PayMeansOfProductionForm):
         data = self.controller.process_input_data(form)
         if isinstance(data, PayMeansOfProductionController.MalformedInputData):
             self.presenter.present_malformed_data_warnings(data)
-            return Response(self._render_template(), status=400)
+            return Response(self._render_template(form), status=400)
         use_case_response = self.pay_means_of_production(data)
         self.presenter.present(use_case_response)
-        return Response(self._render_template(), status=200)
-
-    def _render_template(self) -> str:
         form = PayMeansOfProductionForm()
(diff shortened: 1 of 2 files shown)


arbeitszeit_flask/views/pay_means_of_production.py
```diff
@@ -18,19 +18,20 @@ class PayMeansOfProductionView:
...
             self.presenter.present_malformed_data_warnings(data)
-            return Response(self._render_template(form), status=400)
+            return Response(self._render_template(), status=400)
         use_case_response = self.pay_means_of_production(data)
         self.presenter.present(use_case_response)
-        return Response(self._render_template(form), status=200)
+        return Response(self._render_template(), status=200)
 
-    def _render_template(self, form: PayMeansOfProductionForm) -> str:
+    def _render_template(self) -> str:
```
seppeljordan, line 33: With this approach the form is cleared after every request. This introduces a new (minor) problem. If the user makes an error when filling the form and submits it, they get an empty form back. Let's say the user forgot to copy the last character of the plan id. When they sumbit the form, the application will complain that the user input data was incorrect. But instead of keeping the fields like amount and purpose filled in the form will be empty and the user as to fill those fields again.

Consider only clearing the form when the request was handled successfully. Another approach would be to redirect to this instead of responding directly (in line 31).

seppeljordan (changes requested):

arbeitszeit_flask/views/pay_means_of_production.py
sloschert: Good point. I went for the first approach.

seppeljordan (approved):

seppeljordan:
Certs transferred

PR #378, opened 2022-05-15
Implement (non-functional) /accountant/signup/<token> route

seppeljordan:
This PR implements the basis of a web view for the register accountant view. It does not work currently since there is no proper controller and presenter in place to handle requests. This is planned for a future PR though. I tried to keep this PR really small to reduce the review burden.

I'd argue that it is safe to merge this PR since the view does currently not do anything. The next PR will implement the presenter and controller to handle proper registration requests.


sloschert:
There is an empty file `register_accountant_view.py`. I guess it can be deleted?

sloschert (approved):
Great! I had one comment. Certs transferred.

PR #383, opened 2022-05-20
Implement more accountant registration logic

seppeljordan:
4 things were done:

# Implement LogInAccountant use case

The use case should mainly just verify users credentials. The the actual user credential validation is still done in flask. The `AccountantRepository` is expected to provided a method `validate_credentials` that signals if credentials provided by the user are valid.

# Implement GetAccountantProfileInfo use case

The use case is analogous to the GetMemberProfileInfo use case.

# Improve RegisterAccountant use case

The business logic for accountant registration should be more or less complet.

# Implement AccountantGenerator for testing

The newly implemented `AccountantGenerator` differs a bit from the other two user generator classes in that it does not return an entity to the caller but only the uuid. The idea here was that the tests should not use any of the entities since they are supposed to be internal to the use cases. Instead the testing should only be done in terms of observable behavior. With observable behavior I mean use case response object.

### DIFF
--- a/tests/use_cases/test_get_accountant_profile_info.py
+++ b/tests/use_cases/test_get_accountant_profile_info.py
@@ -16,7 +16,7 @@ class UseCaseTests(TestCase):
         self.get_profile_use_case = self.injector.get(GetAccountantProfileInfoUseCase)
         self.accountant_generator = self.injector.get(AccountantGenerator)
 
-    def test_that_user_user_can_see_correct_email_if_they_registered_with_test_mail_address(
+    def test_that_user_can_see_correct_email_if_they_registered_with_test_mail_address(
         self,
     ) -> None:
         email_address = "test@test.test"
@@ -28,7 +28,7 @@ class UseCaseTests(TestCase):
         )
         self.assertRecord(response, lambda r: r.email_address == email_address)
 
-    def test_that_user_user_can_see_correct_email_if_they_registered_with_another_test_mail_address(
+    def test_that_user_can_see_correct_email_if_they_registered_with_another_test_mail_address(
         self,
     ) -> None:
         email_address = "test2@test.test"


arbeitszeit/use_cases/register_accountant.py
```diff
@@ -13,13 +20,34 @@ class Request:
...
+    member_repository: MemberRepository
+    company_repository: CompanyRepository
+    accountant_repository: AccountantRepository
 
     def register_accountant(self, request: Request) -> Response:
         invited_email = self.token_service.unwrap_invitation_token(request.token)
-        return self.Response(is_accepted=invited_email == request.email)
+        if invited_email != request.email:
+            return self._failed_registration()
+        if self.member_repository.has_member_with_email(request.email):
```
sloschert, line 39: this makes that an accountant cannot use the same mail adress he/she uses for their company or member account. is this restriction necessary?

tests/use_cases/test_get_accountant_profile_info.py
```diff
@@ -0,0 +1,76 @@
...
+        email_address = "test@test.test"
+        user_id = self.accountant_generator.create_accountant(
+            email_address=email_address
+        )
+        response = self.get_profile_use_case.get_accountant_profile_info(
+            request=GetAccountantProfileInfoUseCase.Request(user_id=user_id)
+        )
+        self.assertRecord(response, lambda r: r.email_address == email_address)
+
+    def test_that_user_user_can_see_correct_email_if_they_registered_with_another_test_mail_address(
```
sloschert, line 31: why is this necessary? i thought each test function gets a fresh setup anyway.

sloschert (approved):
Nice. Regarding the "only one email-adress per user"-restriction: Currently I don't see neither why we would want this restriction. I'd leave it up to you if you want to revert this restriction right away or discuss/think about it once more. 

Certs already transferred.

sloschert (approved):
Thanks.

PR #385, opened 2022-05-21
Reduce code duplication in templates

seppeljordan:
The goal of the PR was to streamline the templating in the app.

I refactored the `base.html` template to include two basic blocks: The `navigation` block and the `content` block. All other templates should follow this basic structure. With this change the `base_member.html` and `base_company.html` can simply override the `navigation` block. Also the `unconfirmed_member.html` and `unconfirmed_company.html` no derive from their respective user types base template `base_member.html` and `base_company.html`.

As a consequence the `block member` and `block company` directives both got deleted. Instead we use `block content` everywhere.

### DIFF
--- a/arbeitszeit_flask/templates/auth/login_company.html
+++ b/arbeitszeit_flask/templates/auth/login_company.html
@@ -1,6 +1,6 @@
 {% extends "base_company.html" %}
 
-{% block company %}
+{% block content %}
 <div class="section is-medium">
     <div class="container">
         <div class="column is-4 is-offset-4">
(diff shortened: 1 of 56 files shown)


sloschert:
Create change. There are only 2 things on the start page (`start.html`) that I would prefer to have changed:

- entry/welcome text is not vertically centered (originally it was done with a bulma hero section, there might be other ways to achieve it, too https://bulma.io/documentation/layout/hero/)
- there is a new "start" button in the right corner of the navbar that has no effect

sloschert (approved):
Thanks! Certs transferred.

PR #390, opened 2022-05-23
Implement registration web logic for accountants

seppeljordan:
This PR implements proper registration for accountants.

# Walk through

I want to describe and show how the current registration system is implemented. It is not perfect by any means but I am confident that it is serviceable for now.

First an administrator having access to the application on the server via the shell has to issue an invitation:

```
$ flask invite-accountant jan-appel@gik.org
```

This will send an email to the specified address containing an invitation link. Here is the output in development mode:

```
$ flask invite-accountant jan-appel@gik.org
Email would be sent:
recipients: jan-appel@gik.org
subject: Invitation to Arbeitszeitapp
sender: None
content: <p>
  Hello. You were invited to be a public accountant at Arbeitszeitapp. Go to <a href="http://localhost:5000/accountant/signup/Imphbi1hcHBlbEBnaWsub3JnIg.YoudlQ.ruHC3ZYWwIY_1tpefF24_lHEx0E">http://localhost:5000/accountant/signup/Imphbi1hcHBlbEBnaWsub3JnIg.YoudlQ.ruHC3ZYWwIY_1tpefF24_lHEx0E</a> to register.
</p>
```

After clicking entering the link in the browser the user will be greeted by the following view:

accountant-login-screen

The user now has to enter *the same email address as was specified during the invitation*, otherwise the registration will fail.

failed-registration

When the user registers successfully however then they will be logged in and greeted by a blank screen (because dashboard is not implemented yet).

registration-success

# Remaining problems

* We have to discuss if invitation codes should be valid for any email address or just the email address that was specified by the admin.
* Error message have to be improved. Currently the user is only informed about the fact that a registration has failed but they are not given any explanation about the reason.
* The admin command needs to give feedback if an invitation failed (e.g. because the email address already belongs to an accountant)

### DIFF
--- a/arbeitszeit_flask/forms.py
+++ b/arbeitszeit_flask/forms.py
@@ -108,25 +108,23 @@ class RegisterAccountantForm(Form):
         trans.lazy_gettext("Email"),
         validators=[
             validators.InputRequired(
-                message=trans.lazy_gettext("Emailadresse erforderlich")
+                message=trans.lazy_gettext("Email address is required")
             )
         ],
     )
     name = StringField(
         trans.lazy_gettext("Name"),
         validators=[
-            validators.InputRequired(
-                message=trans.lazy_gettext("Name ist erforderlich")
-            )
+            validators.InputRequired(message=trans.lazy_gettext("Name is required"))
         ],
     )
     password = PasswordField(
-        trans.lazy_gettext("Passwort"),
+        trans.lazy_gettext("Password"),
         validators=[
             validators.Length(
                 min=8,
                 message=trans.lazy_gettext(
-                    "Passwort muss mindestens 8 Zeichen umfassen"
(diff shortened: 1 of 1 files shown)


arbeitszeit_flask/forms.py
```diff
@@ -103,6 +103,34 @@ def add_email_error(self, error: str) -> None:
         self.email.errors.append(error)
 
 
+class RegisterAccountantForm(Form):
+    email = StringField(
+        "Email",
+        validators=[validators.InputRequired(message="Emailadresse erforderlich")],
```
sloschert, line 109: should be translated

sloschert (approved):
Translation of strings to English would be great ;-)  
Apart from that superb. Certs transferred.

PR #392, opened 2022-05-23
Refactor DB plan table and plan entity

seppeljordan:
This PR removed the `.approved` field of `arbeitszeit_flask.models.Plan` since the approval state of a plan can be deduced from `approval_date`. Further I removed the `.approved` attribute of the `arbeitszeit.entities.Plan` and replaced it with the `is_approved` property which simply checks if the `approval_date` is set.

I also removed an "illegal" test that checked for `AssertionError`. IMO assertions should not exist caught or tested. Instead assertions should only be there for the programmer to clarify some aspect of the code or as a last resort to prevent illegal state in the application.

### DIFF
--- a/arbeitszeit/entities.py
+++ b/arbeitszeit/entities.py
@@ -134,7 +134,6 @@ class Plan:
     description: str
     timeframe: int
     is_public_service: bool
-    approved: bool
     approval_date: Optional[datetime]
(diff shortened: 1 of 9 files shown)


sloschert (approved):
Certs transferred.

PR #394, opened 2022-05-24
Simplify configuration

seppeljordan:
The principle of this refactoring is to simplify the configuration of
the application. A canonical configuration file location was
established at /etc/arbeitszeitapp/arbeitszeitapp.py.  The
ARBEITSZEITAPP_CONFIGURATION_PATH allows to specify an alternate
configuration file location.

The file arbeitszeit_flask/configuration_base.py now is a location for
default settings. Instead of importing the file from other settings
files it is automatically loaded on application startup. All
configuration options from this file can be overridden by the user
specified configuration.

The documentation on app config was extended.

### DIFF
--- a/arbeitszeit/use_cases/register_accountant.py
+++ b/arbeitszeit/use_cases/register_accountant.py
@@ -22,6 +22,7 @@ class RegisterAccountantUseCase:
     class Response:
         is_accepted: bool
         user_id: Optional[UUID]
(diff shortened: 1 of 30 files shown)


sloschert (approved):
top. certs transferred to plan ID 1906a785-a23b-44b3-a599-96ee43dd388c.

PR #397, opened 2022-05-25
Refactor /company/seek_approval

seppeljordan:
A presenter was introduce and a obsolete template
/templates/company/create_plan_response.html was removed.

When filing a plan the application redirects to the companies plan overview page /company/my_plans since the create_plan_response.html template, which was used to render the response to /company/seek_approval, did not provide any information.

EDIT: The testing done on the presenter is not as throughout as usual since most of it will get deleted when implementing proper plan inspection anyway.

### DIFF
--- a/arbeitszeit_flask/company/routes.py
+++ b/arbeitszeit_flask/company/routes.py
@@ -1,7 +1,7 @@
 from typing import Optional
 from uuid import UUID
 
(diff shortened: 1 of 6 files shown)


sloschert:
Certs transferred.

sloschert (approved):

PR #400, opened 2022-05-25
Allow configuration of https policy

seppeljordan:
This PR allows administrators to configure the handling of insecure connections via HTTP.  The admin can now specify `FORCE_HTTPS=False` in the settings file to disable force connection upgrades to HTTPS. This is useful for automatic testing of deployment configurations. The default value for this option is ``True``, meaning that the admin has to make the conscious decision to allow HTTP.


sloschert (approved):
Certs transferred.

PR #401, opened 2022-05-25
Include arbeitszeit_flask/static/**/*.{css,js} in package distributions

seppeljordan:
Previously the static files were not included in package distributions of the `arbeitszeitapp`. By including javascript and css files in the `arbeitszeit_flask/static` in package distribution users can install the `arbeitszeitapp` from a python package instead of having to clone the code directly to their production server.


sloschert (approved):
Certs transferred

PR #403, opened 2022-05-25
Package migrations with arbeitszeit_flask application

seppeljordan:
This PR moves the migrations to the `arbeitszeit_flask` package. This enables us to package them as part of the application. This change should require no action on the production server. It should all "just work".


sloschert (approved):
Certs transferred.

PR #407, opened 2022-05-28
Implemen auto migrate

seppeljordan:
This PR enables a "auto migrate" feature. If the `AUTO_MIGRATE` setting is `True` then the application will migrate the database automatically on startup. This is useful for production server scenarios since it reduces the management work that has to be done by the admin.


sloschert (approved):
Certs transferred.

PR #410, opened 2022-05-29
Refactor /company/transfer_to_company

seppeljordan:
This refactoring tried to make the whole form handling easier to deal
with. This refactoring is based on some assumptions on how form
handling should be done.

1) When a form is successfully processed the user should be
immediately redirected.

2) It is the presenters responsibility to decide whether to redirect
or not.

3) If the user is not redirected then the form was not processed
successfully.

4) Conversion of strings from the web form should be done the the form
implementation itself.

As a result I deleted tests and error handling code from the
controller. I think that the result is easier to understand and reason
about.


sloschert (approved):
Certs transferred.

PR #420, opened 2022-06-05
Configure babel setuptools integration

seppeljordan:
This change was done to increase compatibility with python standard
tools. With this change the "old" way of handling translations was
dropped. The "new" way to deal with translations was documented in
README.md.


sloschert (approved):
Good! Certs transferred.

PR #421, opened 2022-06-06
Package translations when building wheel

seppeljordan:
With this setup the compiled translations (.mo files) will be in included in wheels (binary distributions) of arbeitszeitapp. This is beneficial since it makes sure the compiled translations are available for users without commiting binary files to the source code repository.


sloschert (approved):
Certs wanted?

Issue #428, opened 2022-06-10
Prices of past purchases of companies are not rounded


sloschert:
In Company-route `company/kaeufe` (endpoint: `arbeitszeit_flask/company/routes.py` -> `my_purchases`) the prices of past purchases of members are not rounded. Specifically, `price_per_unit` and `price_total` are not rounded.

It looks like this: 
![image](https://user-images.githubusercontent.com/61537351/173097042-6f2b722f-8792-423a-af82-75d1c704688b.png)

Actually this is not the only problem, there is some refactoring pending (create controller, presenter, view object plus tests). For example, right now the use case is called directly in the route and the use case result is directly passed into the html template, without a presenter involved. This is far from ideal.

But anyway, it would be good to resolve the rounding-issue first (also to gain experience).

I see two ways to do this without doing all the refactoring first: 1) Round both attributes directly in the use case (and write tests for this rounding functionality in `tests/use_cases/test_query_purchases.py`) 2) Round both attributes in the jinja template (https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.round)

jacknely:
Hi, I'll work on this.

PR #429, opened 2022-06-10
fix(#428): ROUND PAST PRICE OF COMPANIES

jacknely:
Update jinja2 template to round `price_per_unit` and `price_total` to 1 decimal place.

https://github.com/arbeitszeit/arbeitszeitapp/issues/428


sloschert (approved):
Thanks for contributing, looks good to me!

PR #430, opened 2022-06-11
Implement use case for changing the app language

seppeljordan:
This PR refactors the changing of app language. No functional changes were made. With this patch I wanted to prepare a more extensive overhaul of the system since it is currently really limited.


sloschert (approved):
Great work! Certs transferred.

PR #437, opened 2022-06-19
Implement LogInCompanyPresenter

seppeljordan:
This is a refactoring. The behavior of the login view for companies should not change with this PR.


sloschert (approved):
Top! Certs transferred.

PR #443, opened 2022-06-24
Refactor dependency injection

seppeljordan:
This PR moves all providers for presenters into their own module at `arbeitszeit_flask.dependency_injection.presenters`. This change was done because the length of the file `arbeitszeit_flask/dependency_injection/__init__.py` was getting excessively large.

This should also provide clear orientation for programmers on where to find the dependency injection directives they are looking for.

EDIT: Also I noticed that there were 2 different `PlanSummaryService` protocols. This was rather confusing to me so I renamed one of them to `PlanSummaryFormatter` since this was its actual function.


sloschert (approved):
Thanks. Certs transferred.

PR #444, opened 2022-06-24
Update nix and python dependencies

seppeljordan:
Update of the build environment. It is recommended that developers update the virtual env on their development machine if `pip` is used to manage the development environment.

Also I updated `requirements.txt` file setup. I split `requirements.txt` into three separate files: `requirements.txt`, `requiements-dev.txt` and `constraints.txt`. The "new" `requirements.txt` file contains only the package names of **direct dependencies**. No version numbers specified. Instead it includes the file `contraints.txt` file via the `-c` directive. `requirements-dev.txt` contains all dependencies that are required to run the tests and linting.

`constraints.txt` contains the package names and version strings for **all direct and indirect dependencies** of `arbeitszeitapp` as found in the nix environment. If pip encounters a package dependency when installing the dependencies it will always pick the version as specified in `contraints.txt`. Packages that are not needed for production, like for example `flake8` can remain in `constraints.txt` without being installed on the production environment since packages from `containts.txt` are only installed if they are explicitly required by deps found in `requirements.txt`.

Due to some dependency conflicts I had to exclude `mccabe` and `docutils` from `constraints.txt`. This means that pip will pick a version for these packages when running `pip install`. This should not be a big deal since those two are only required for linting and building the docs.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,4 +7,5 @@ hypothesis
 isort
 mypy
 psycopg2
+pytest
 sphinx


sloschert:
So for developers who use pip it's recommended running `requirements-dev.txt`, right? Beause if running `requirements.txt`, flake8, mypy, etc won't get updated.

sloschert (approved):
Thanks for the efforts, certs transferred.

PR #445, opened 2022-06-24
Remove expiration_relative field from Plan entity

seppeljordan:
This PR removes the column `expiration_relative` from the database model `Plan` and the `expiration_relative` attribute from the entity `Plan`. The reasoning is that the `expiration_relative` day count can always be calculated on demand from the absolute `expiration_date` field and the current time.  To retain the current functionality the calculation logic for the remaining days of an active `Plan` was moved to the presenter layer.

This changed allowed me to remove unnecessary database operations.

### DIFF
--- a/tests/presenters/test_show_my_plans_presenter.py
+++ b/tests/presenters/test_show_my_plans_presenter.py
@@ -4,8 +4,10 @@ from unittest import TestCase
 
 from arbeitszeit.entities import Plan
 from arbeitszeit.use_cases.show_my_plans import PlanInfo, ShowMyPlansResponse
+from arbeitszeit.use_cases.update_plans_and_payout import UpdatePlansAndPayout
 from arbeitszeit_web.show_my_plans import ShowMyPlansPresenter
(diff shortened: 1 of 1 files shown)


tests/presenters/test_show_my_plans_presenter.py
```diff
@@ -12,61 +12,6 @@
...
-    plan_info = _convert_into_plan_info(plan)
-    return ShowMyPlansResponse(
-        count_all_plans=1,
-        non_active_plans=[plan_info],
-        active_plans=[],
-        expired_plans=[],
-    )
-
-
 class ShowMyPlansPresenterTests(TestCase):
```
sloschert, line 15: Probably we should test the corect calculation/display of the ViewModel's `expiration_relative` attribute?

sloschert: strange. I was just referring to `class ShowMyPlansPresenterTests(TestCase)`

sloschert (approved):
Certs transferred.

PR #447, opened 2022-06-29
mypy: Don't check site-packages

seppeljordan:
This change directs `mypy` not to check installed "site-packages", meaning installed third party packages.

This improves mypy runtime significantly without compromising type safety in the project.

No cert transfer needed.


sloschert (approved):
great.

PR #448, opened 2022-06-30
Refactor /member/confirm/<token>

seppeljordan:
This PR makes an attempt in removing random DB access from the routes by implementing proper methods on the repository. This is also preparation work for moving the confirmation logic into the business logic and forming a use case around it.

### DIFF
--- a/tests/data_generators.py
+++ b/tests/data_generators.py
@@ -70,10 +70,7 @@ class MemberGenerator:
         account: Optional[Account] = None,
         password: str = "password",
         registered_on: Optional[datetime] = None,
-        is_confirmed: Optional[bool] = None,
     ) -> Member:
-        if is_confirmed is None:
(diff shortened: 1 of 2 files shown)


tests/data_generators.py
```diff
@@ -69,8 +69,11 @@ def create_member(
         name: str = "test member name",
         account: Optional[Account] = None,
         password: str = "password",
-        registered_on: datetime = None,
+        registered_on: Optional[datetime] = None,
+        is_confirmed: Optional[bool] = None,
     ) -> Member:
+        if is_confirmed is None:
```
sloschert, line 75: `is_confirmed` is not used in this method. are you planning to do something with this later?

sloschert:
I have one remark. The rest looks good.

sloschert (approved):
Certs transferred.

PR #449, opened 2022-07-01
Implement CompanyRepository methods for company confirmation

seppeljordan:
I implemented `CompanyRepository.confirm` and `CompanyRepository.is_confirmed` to move away from using models directly in views. This is part of a move towards implementing proper use cases for user confirmation.

EDIT: I took the liberty to move all instances of confirmation handling to the respective repositories.


sloschert (approved):
Certs transferred

PR #457, opened 2022-07-06
Rename SeekApproval -> SelfApprovePlan

seppeljordan:
This PR renames the use case, presenter and associated tests around `SeekApproval` to `SelfApprovePlan`. This is the first step in the refactoring process. The next step will be to incorporate the `ActivatePlanAndGrantCredit` use case into `SelfApprovePlan`.

EDIT: I also regenerated the API docs via the `./regenerate-api-docs` command.


sloschert (approved):
Certs transferred.

PR #460, opened 2022-07-09
Remove ActivatePlanAndGrantCredit use case

seppeljordan:
This PR gets rid of the `ActivatePlanAndGrantCredit` use case. All this use case did is not part of `SelfApprovePlanUseCase`.

As a consequence many error states that were previously possible are now impossible. Especially everything that deals with approved but inactive plans cannot be tested anymore. I adjusted all tests that were broken but still made sense is essence. Some tests got deleted though since there was no way of reproducing the situations that those tests were testing.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,55 @@
-Development setup
-=================
+Overview
+========
+
+Description
+-----------
+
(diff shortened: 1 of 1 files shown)


sloschert (approved):
Thanks, that was quite some work. Certs transferred.

PR #466, opened 2022-07-21
Refactor PlanGenerator to use CreatePlanDraft use case instead of repository

seppeljordan:
This PR refactors the `PlanGenerator` used in tests to generate plan drafts through the proper use case instead of the `DraftRepository`. I think that this step is necessary lest we run the risk of duplicating business logic in our test setup.


sloschert (approved):
:+1: Certs transferred

PR #469, opened 2022-07-22
Remove docker image from nix flake

seppeljordan:
This PR removes the docker image from the nix flake.

I think that it is appropriate to remove the docker image since it was not used in anyway and not properly functional. Having this in the code might make people believe that the image worked. I think that we should discuss providing docker images at a later point, e.g. when the application has a stable release.

No certificates needed.


sloschert (approved):
agree!

PR #471, opened 2022-07-24
Implement EditDraftUseCase

seppeljordan:
This PR implements a `EditDraftUseCase`. This use case provides the business logic for companies to edit their drafts. There is currently no web code implemented for this use case but it is properly tested. The follow up to this PR will include the necessary web logic.

EDIT: I tried something new the the repository interface. The new `update_draft` method now takes a "update model" as an argument. This model describes which attributes should be updated on the draft. I think that this makes for a more flexible approach to the database interfaces. Let me know what you think of this approach.


sloschert (approved):
Great PR. I think the update model is a good innovation. Certs transferred.

PR #472, opened 2022-07-24
Refactor tests.controllers to use depenency injection framework

seppeljordan:
This PR refactors all the controller tests to utilize a proper dependency injector instead of constructing classes "manually".


sloschert (approved):
:+1: C.T.

PR #474, opened 2022-07-26
Write documentation on how we develop database interfaces

seppeljordan:
I wrote some documentation on how to implement database interface code (aka repositories). I tried to illustrate the explanations given by example code. The example code is for a fictional use case. This was done on purpose so that the reader can focus on the coding strategies instead of the complex domain of the Arbeitszeitrechnung.

Here is a link to file how it is on the git branch for the PR. I think it is easier to read through the proposed new documentation over there instead of reading the raw diff output:
https://github.com/arbeitszeit/arbeitszeitapp/blob/seppeljordan/document-repository-interfaces/docs/philosophy.rst#implementing-calls-to-persistent-storage-aka-the-database

### DIFF
--- a/docs/philosophy.rst
+++ b/docs/philosophy.rst
@@ -152,7 +152,7 @@ protocol where the data about the books is stored in an SQL database::
 Now let us see how another class can use the interface to implement a
 use case::
 
-  # Implement a use case makes of the BookStorage protocol. This class
+  # Implement a use case that uses the BookStorage protocol. This class
   # implements the "business logic" of borrowing books from a library.
   # The customer cannot borrow any more books if they currently borrow
   # 20 other books.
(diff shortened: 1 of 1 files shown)


docs/philosophy.rst
```diff
@@ -66,13 +66,168 @@ See this example for reference::
...
+          )
+
+      def run_query(self, query, *args):
+          # run the specified query against the database
+          pass
+
+Now let us see how another class can use the interface to implement a
+use case::
+
+  # Implement a use case makes of the BookStorage protocol. This class
```
sloschert, line 155: "Implement a use case makes" - typo?

docs/philosophy.rst
```diff
@@ -66,13 +66,168 @@ See this example for reference::
...
+          customer_id = update.book_ids
+          for book in update.book_ids:
+              self.books[book] = customer_id
+
+      def create_book(self, book_id: int, borrowed_to: Optional[int]) -> None:
+          self.books[book_id] = borrowed_to
+
+The protocols that define DB access in the *arbeitszeitapp* are
+located under ``arbeitszeit.repositories``. Note that while the above
+method is the standard for new implementations there is some "legacy"
```
sloschert, line 229: a suggestion, to make clearer, what the new standard is: "Not that while the above method, *using protocol*, is the standard for ..." 
?

sloschert (approved):
well explained. CT

sloschert:
> I included your suggestions. 

:+1:

PR #486, opened 2022-08-07
Simplify UrlIndex setup

seppeljordan:
This PR aims to simplify the url index setup. The mid-term goal is to simplify the url index setup down to basically one interface `arbeitszeit_web.url_index.UrlIndex` and two implementation `arbeitszeit_flask.url_index.GeneralUrlIndex` and `tests.presenter.url_index.UrlIndexImpl`.

I refactored all presenters relying on the old `PlanSummaryUrlIndex` in a way that they use the newly created `UrlIndex` interface where separate methods for getting plan summary urls are declared: `get_company_plan_summary_url` and `get_member_plan_summary_url`. This allows for much easier injection logic. This lead to a drastic reduction in dependency injection related code.

I tried to document the ideas behind the setup in `arbeitszeit_web.url_index` and provide direction for developers on how to use `UrlIndex` interface and the new context aware `UserUrlIndex`.

Please let me know what you think about this change.

### DIFF
--- a/arbeitszeit_web/get_coop_summary.py
+++ b/arbeitszeit_web/get_coop_summary.py
@@ -5,8 +5,9 @@ from typing import Any, Dict, List
 from injector import inject
 
 from arbeitszeit.use_cases.get_coop_summary import GetCoopSummarySuccess
+from arbeitszeit_web.session import Session
 
-from .url_index import EndCoopUrlIndex, UserUrlIndex
+from .url_index import EndCoopUrlIndex, UrlIndex, UserUrlIndex
 
 
 @dataclass
@@ -25,6 +26,8 @@ class GetCoopSummaryViewModel:
     coop_name: str
     coop_definition: List[str]
     coordinator_id: str
+    coordinator_name: str
+    coordinator_url: str
 
     plans: List[AssociatedPlanPresentation]
 
@@ -37,6 +40,8 @@ class GetCoopSummaryViewModel:
 class GetCoopSummarySuccessPresenter:
     user_url_index: UserUrlIndex
     end_coop_url_index: EndCoopUrlIndex
+    url_index: UrlIndex
+    session: Session
 
     def present(self, response: GetCoopSummarySuccess) -> GetCoopSummaryViewModel:
         return GetCoopSummaryViewModel(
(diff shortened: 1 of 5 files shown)


arbeitszeit_web/url_index.py
```diff
@@ -120,3 +134,27 @@ def get_accountant_dashboard_url(self) -> str:
...
+
+
+@inject
+@dataclass
+class UserUrlIndex:
+    """This class is not an interface and therefore should not be
+    implemented by the web framework. It is merely used internally as
+    a convinience interface. You should refrain from using this class
+    in your tests and instead rely on the UrlIndex interface. In a
+    test scenario you most likely now in advance if you expect a url
```
sloschert, line 146: know statt now

arbeitszeit_flask/dependency_injection/__init__.py
```diff
@@ -282,6 +263,14 @@ def provide_prefilled_draft_data_controller(
 
 
 class FlaskModule(PresenterModule):
+    @provider
+    def provide_request(self, request: FlaskRequest) -> Request:
+        return request
+
+    @provider
+    def provide_plan_summary_url_index(self, index: GeneralUrlIndex) -> UrlIndex:
```
sloschert, line 271: why is this called `provide_plan_summary_url_index`? isn't its purpose more general?

arbeitszeit_web/get_coop_summary.py
```diff
@@ -30,10 +33,12 @@ def to_dict(self) -> Dict[str, Any]:
         return asdict(self)
 
 
+@inject
 @dataclass
 class GetCoopSummarySuccessPresenter:
-    plan_url_index: PlanSummaryUrlIndex
+    user_url_index: UserUrlIndex
     end_coop_url_index: EndCoopUrlIndex
+    session: Session
```
sloschert, line 41: as far as I see session is not needed

tests/presenters/dependency_injection.py
```diff
@@ -132,22 +153,18 @@ def provide_toggle_plan_availability_url_index(
 
     @singleton
     @provider
-    def provide_plan_summary_url_index_test_impl(self) -> PlanSummaryUrlIndexTestImpl:
-        return PlanSummaryUrlIndexTestImpl()
+    def provide_plan_summary_url_index_test_impl(self) -> UrlIndexTestImpl:
+        return UrlIndexTestImpl()
+
+    @provider
+    def provide_plan_summary_url_index(self, index: UrlIndexTestImpl) -> UrlIndex:
```
sloschert, line 160: same question here, why that name...

sloschert (approved):
I like the change, I only have one question.
Certs already transfered though.

sloschert:
Looks good. 
As a side note: Actually I was always thinking that pages like company summary or cooperation summary may look exactly the same, no matter if a member, a company or an accountant looks at it. But 1) we don't really know if we want that to be the case in the future and 2) we have these pages right know separated in member/company routes. 
So I guess the hassle with the different UserRoles is just necessary.

PR #487, opened 2022-08-14
Simplify UrlIndex

seppeljordan:
This PR tries to reduce the complexity in working with `UrlIndex` and adjacent classes. The `PlotsUrlindex` protocol got completely removed and its methods merged into the `Urlindex` protocol. Furthermore the `AnswerCompanyWorkInviteUrlIndex` and `CompanyUrlIndex` protocols got removed and their methods merged into `UrlIndex`. The two methods `get_member_plan_summary_url` and `get_company_plan_summary_url` were merged into a single method `get_plan_summary_url(self, user_role: Optional[UserRole], plan_id: UUID) -> str` to make it consistent with other methods on the `Urlindex` protocol that are user role aware.


sloschert (approved):
:+1: CT

PR #493, opened 2022-08-21
Show payout factor (FIK) on statistics page

### LINKED ISSUE #245: Show FIK to users
Currently FIK (factor individual consumption) is calculated implicitly in a private function (`_calculate_payout_factor`), but is shown nowhere to members or companies. 

Possible places to show FIK are

- Statistics
- Account overview (where wages are paid out)
- Profile page of members

sloschert:
fixes #245

This PR shows the payout factor ("FIK") on the statistics page, as discussed in issue #245.

A new `PayoutFactorRepository` was created.

Also a new entity `PayoutFactor` was created, that includes apart from the value a timestamp. I considered to create just a data transfer object, but then decided that the payout factor deserved an entity object and that it's valid to include a timestamp. I hope that makes sense. 

Plan-ID: ed6eb89f-0633-4544-9ddc-94466b7d5e82

Commits: 

- show payout factor timestamp on stats page
- show current payout factor on statistics page
- move storing of payout fctr in PayoutFactorService
- db migration file to create payout factor table
- let update_plans_and_payout store payout factor
- introduce payout factor repository

Screenshot:

Bildschirmfoto vom 2022-08-21 13-32-10


sloschert:
Apparently `nix flake check` is failing, because it checks on python 3.9, while I'm using the bisect.insort-parameter `key` which is quite new.

@seppeljordan What do you think, should I find a workaround in the code or should we make a change to `flake.nix`? Not sure how important it is to check on python 3.9

sloschert:
How is it done? I tried to just delete the lines with python39, but that's not working, I can't figure out why.

PR #496, opened 2022-08-28
Show drafts in "my plans" overview

seppeljordan:
With this change all drafts that a company has created are now displayed in the "my plans" overview of that company. This is a jumping off point for creating a proper management interface for drafts.

In a future PR we should implement edit and delete buttons for those drafts.


sloschert (approved):
Great. CT.

PR #497, opened 2022-08-31
Implement DeleteDraft use case

seppeljordan:
This PR implements the business logic for deleting plan drafts. It is implemented such that only the planner themself can delete a draft. In a follow up PR I want to implement the web logic for deleting plans.


sloschert (approved):
CT

PR #515, opened 2022-09-24
Implement proper button to delete plan drafts

seppeljordan:
This PR creates a button that lets the user delete their drafts if so desired. The responsible view `company_main.delete_draft` will redirect to the view that the button was clicked on.

The styles on the draft overview page `/company/draft` looks a little janky but I refrained from putting to much work into fixing this issue. We will probably reimplement the html template for this page very soon anyway.


sloschert (approved):
Great PR.
CT

PR #516, opened 2022-09-24
Create FilePlanWithAccounting use case

seppeljordan:
This is the business logic part for filing plans. This PR also creates a stub use case ListPlansWithPendingReview. This use case is not fully functional. However the parts that are already there are properly tested.

Further there are also presenter and controller for the FilePlanWithAccounting use case.

While working on this issue I realized that our data model does not allow us to model rejected plans. As a next step I would like to rework our models around plan approval/rejection.


sloschert (approved):
Great. CT.

PR #522, opened 2022-09-27
PayConsumerProduct: Account for the case that buyer is not valid

seppeljordan:
I wanted to remove the `assert` statement in PayConsumerProduct use case. This PR implements all the logic for that.


sloschert (approved):
👍 CT

PR #524, opened 2022-09-28
Document subclassing unittest.TestCase

seppeljordan:
This PR changes the documentation. It explains how to properly subclass unittest.TestCase.

No certs needed.


sloschert (approved):
Good explanation.

PR #526, opened 2022-09-29
Remove hypothesis from dependencies

seppeljordan:
I removed the test dependencies `hypothesis`. It was only used in one test case and nowhere else. I think that the additional overhead of having `hypothesis` around (compile time, maintaining compatability) is not worth it.

No certs needed.


sloschert (approved):
D'accord!

PR #528, opened 2022-09-29
Philip/tests for windows

### LINKED ISSUE #495: Module tests do not run on Windows
**(1) Error with year '1'**
In file _show_my_plans.py_:
```
def __format_date(self, date: Optional[datetime]) -> str:
      return f"{date.strftime('%d.%m.%y')}" if date else "-"
      ValueError: format %y requires year >= 1900 on Windows
```
Called with Pythons  datetime.MINYEAR == 1 (generated by _datetime.min_) in _test_show_my_plans_presenter.py_ , Line 52 and 61

```
 plan = self.plan_generator.create_plan(activation_date=datetime.min)
```

**(2) 2nd error: AssertionError**
```
E       AssertionError: assert <TransactionTypes.payment_of_liquid_means: 7> == <TransactionTypes.sale_of_fixed_means: 1
1>
E        +  where <TransactionTypes.payment_of_liquid_means: 7> = TransactionInfo(transaction_type=<TransactionTypes.pay
ment_of_liquid_means: 7>, date=datetime.datetime(2022, 8, 24, 20, 59, 35, 422251), transaction_volume=Decimal('-10'), ac
count_type=<AccountTypes.r: 'r'>, purpose='test purpose').transaction_type
E        +  and   <TransactionTypes.sale_of_fixed_means: 11> = TransactionTypes.sale_of_fixed_means

tests\use_cases\test_get_company_transactions.py:203: AssertionError
```

There are several assert errors. They all result from a wrong/ not expected order of transactions.
These action have been created with the `TransactionGenerator` and sorted with `GetCompanyTransactions`. The `TransactionGenerator` uses the creation `time now_minus_one_day()` for all new transactions. 

The problem is an isufficient time resolution of the `now()` Function on Windows machines. So, all transactions have the same timestamp (sometimes it is different) and the sort function cannot sort them in the intended order. So the checks fail...

See
https://stackoverflow.com/questions/31352912/python-resolution-of-datetime-now
https://stackoverflow.com/questions/3729169/how-can-i-get-the-windows-system-time-with-millisecond-resolution

philipbroistedt:
Windows test issues fixed
#495


tests/use_cases/test_get_company_transactions.py
```diff
@@ -222,21 +225,27 @@ def test_that_correct_info_for_company_is_generated_in_correct_order_after_sever
     transaction_generator.create_transaction(
         sending_account=company1.means_account,
         receiving_account=company2.product_account,
+        date=transaction_generator.datetime_service.now() - timedelta(hours=5),
```
sloschert, line 228: Here again, you should inject `datetime_service: FakeDatetimeService` and use `datetime_service`.

tests/presenters/test_list_drafts_presenter.py
```diff
@@ -21,7 +21,7 @@ def test(self) -> None:
             results=[
                 ListedDraft(
                     id=draft_id,
-                    creation_date=datetime.min,
+                    creation_date=datetime.now(),
```
sloschert, line 24: In the test module, it's in general preferable to use the `FakeDatetimeService` instead of the built-in `datetime`. See for example `tests/presenters/test_get_company_dashboard_presenter.py`. (But we ourselfs did not always use this consistently in the past)

tests/use_cases/test_get_member_account.py
```diff
@@ -116,18 +117,21 @@ def test_that_correct_info_for_company_is_generated_in_correct_order_after_sever
         sending_account=company1.work_account,
         receiving_account=member.account,
         amount_received=Decimal(12),
+        date=transaction_generator.datetime_service.now() + timedelta(hours=1),
```
sloschert, line 120: Better to inject `FakeDatetimeService`. And since the `GetMemberAccount` Use Case gets transactions from the past it would be probably better if the transactions took place in the past.

sloschert:
Hey Philip, 
thanks for the PR! I had some remark, let me know if you have questions regarding these.

PR #530, opened 2022-09-30
Refactor pytest based tests to class based approach

seppeljordan:
I refactored 4 test modules to use the preferred class based approach. I tried my best not to change the semantics of the code.


sloschert (approved):
:+1: CT

PR #531, opened 2022-09-30
Backport application and tests to python3.9

seppeljordan:
This PR changes some details in the testing code to make the app compatible with python3.9 again. This change is motivated by the fact that python3.9 is still the default python version on a lot systems.

To compensate for the lack of insort I implemented a binary search tree (BST) as replacement. It is a relative primitive implementation of a BST. Later changes could change the implementation to a self balancing tree.

No certs needed.

### DIFF
--- a/tests/use_cases/repositories.py
+++ b/tests/use_cases/repositories.py
@@ -151,7 +151,7 @@ class TransactionRepository(interfaces.TransactionRepository):
 
 
 @singleton
-class CompanyWorkerRepository(interfaces.CompanyWorkerRepository):
+class CompanyWorkerRepository:
     @inject
     def __init__(
         self, company_repository: CompanyRepository, member_repository: MemberRepository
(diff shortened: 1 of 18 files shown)


tests/search_tree.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self) -> None:
+        self.root: Optional[Node[T]] = None
+
+    def insert(self, value: T) -> None:
+        if self.root is None:
+            self.root = Node(value=value)
+            return
+        node = self.root
+        while True:
+            if node.value < value:
```
sloschert, line 24: Aren't the smaller nodes usually placed on the *left* of the current node? "If the value is smaller than the current node, move left" (https://courses.engr.illinois.edu/cs225/sp2022/resources/bst/)

tests/use_cases/repositories.py
```diff
@@ -994,17 +993,33 @@ def get_available_language_codes(self) -> Iterable[str]:
...
 @singleton
 class FakePayoutFactorRepository:
+    @dataclass
+    class _PayoutFactorModel:
+        factor: PayoutFactor
+
+        def __lt__(self, other: FakePayoutFactorRepository._PayoutFactorModel) -> bool:
+            return (
+                self.factor.calculation_date < other.factor.calculation_date
+                and self.factor.value < other.factor.value
```
sloschert, line 1003: why should the factor value play a role here if we want to have a iterable sorted by calculation time?

sloschert (approved):
Thanks!

PR #539, opened 2022-10-05
Translation setup: adapt README, fix bug

sloschert:
For the translation setup there was a step in the README missing (the part with `extract_messages`)

Moreover there was a wrong string interpolation (using translator + f-string) in a test function which prevented the extract_messages command to succeed.

(I'll get a merge commit with #538, but that's ok)

No certs needed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -176,7 +176,7 @@ the command::
 Translation
 -----------
 
-We use `Flask-Babel <https://flask-babel.tkte.ch/>`_ for translation.
+We use `Flask-Babel <https://python-babel.github.io/flask-babel/>`_ for translation.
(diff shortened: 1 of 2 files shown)


philipbroistedt:
"We use `Flask-Babel <https://flask-babel.tkte.ch/>` for translation."

The link above is broken. Maybe you can correct it with this commit.

sloschert:
> https://flask-babel.tkte.ch/

Thanks, I'll do. Then we'll be faster than flask babel themselves ;-) https://github.com/python-babel/flask-babel/issues/205

philipbroistedt (approved):

PR #542, opened 2022-10-28
`BaseTestCase` for use cases + new `DataGenerator.create_member_entity`

seppeljordan:
This PR is tries to improve our testing infrastructure in two ways:

# `tests.use_cases.base_test_case.BaseTestCase`

This class implements a base class for all use case related tests. The `setUp` method provides a `injector.Injector` object so that clients can instantiate the classes under test easily. Further the `BaseTestCase` class uses [descriptors](https://docs.python.org/3/howto/descriptor.html) to implement "on demand" instances for commonly used classes such as different kinds of data generator classes.

# `tests.data_generators.MemberGenerator.create_member`

I changed the above method so that instead of returning an entity it only returns the UUID of the generated member. The goal of this change is to transition our testing strategy to a more use case oriented approach. The old functionality of `create_member` is available under `tests.data_generators.MemberGenerator.create_member_entity`. In case this change gets accepted we need to make sure that no more tests are written using `create_member_entity`.

Please let me know what you think.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -176,7 +176,7 @@ the command::
 Translation
 -----------
 
-We use `Flask-Babel <https://flask-babel.tkte.ch/>`_ for translation.
+We use `Flask-Babel <https://python-babel.github.io/flask-babel/>`_ for translation.
 
 1) Add a new language:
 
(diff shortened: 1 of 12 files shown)


sloschert (approved):
Great work. Certs transferred.

PR #551, opened 2022-11-03
Update dependencies

seppeljordan:
This PR updates the dependencies in `requirements.txt` and in `flake.nix`.

This patch might cause some issues for `nix` users. Since there is a bug in `werkzeug` (which is the tool implementing `flask run`) it is not possible to use `flask run` inside of a nix shell. A workaround is to use `python -m flask run` instead. Before this PR we patched manually the `werkzeug` library. This however caused different problems that are much harder to address then the `flask run` issue. Since `werkzeug` is a build dependency for a lot of programs this patch caused a lot of recompiling when using `nix`.

I personally think that the `python -m flask` workaround is acceptable in the face of the benefits from not patching `werkzeug`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -19,7 +19,6 @@
           packages = {
             default = pkgs.python3.pkgs.arbeitszeitapp;
             inherit (pkgs) python3;
-            nixpkgs = pkgs;
           };
           checks = {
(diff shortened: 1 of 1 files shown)


sloschert (approved):
Ok, I think that makes sense. Certs transferred.

PR #558, opened 2022-11-05
Refactor tests.presenters.url_index.UrlIndexImpl

seppeljordan:
This PR refactors the test implementation of the `UrlIndex`.

In the past the individual methods for the `UrlIndexImpl` were implemented all in the same way: For a given call to a url index method generate a unique string with respect to the method name and the arguments that the method was called with. This approach is error IMO.

It is very easy to copy paste code into this module and produce non-unique strings. Also developers need to be aware the string returned by the test implementation needs to be unique.

The new design addresses this issue by automatically generating these unique strings from the method name and the specified arguments.

One drawback of this new design is that this automatic string comparison does not work for arguments that are sometimes specified as kwargs and other times as args. But since we should only use kwargs with the url generator anyway (for clarity) this is okay IMO.

Please let me know what you think about this.


sloschert (approved):
I think that's a good innovation! CT

Issue #555, opened 2022-11-04
Better user experience for companies filing a plan needed


sloschert:
For companies: The process of creating a plan draft until filing that draft for review should be better guided. Until now it is not very intuitive for companies what is going on and what to do next.

At least the newly created draft should be on top of the list of created drafts.

sloschert:
Closing this issue even though it is not resolved. Reasoning:

- We should at first agree on requirements or at least a clearer problem definition. 
- The issue is too broad.

PR #559, opened 2022-11-06
Show drafts ordered; Create "drafts" html anchor

### LINKED ISSUE #532: Plan drafts should be ordered by date, descending
Currently a company's own plan drafts are shown in two different places: In `/company/draft` and in `/company/my_plans`.

In both places, plan drafts are ordered by creation date ascending while the order should be descending. 

This should probably be changed in the two presenters respectively:
- arbeitszeit_web/list_drafts_of_company.py
- arbeitszeit_web/show_my_plans.py

sloschert:
Fixes #532
Related to #555

This PR changes some  things in the drafts list of companies, that can be found under `/company/my_plans`:

- It adds new functionalities to the drafts list (edit button, creation date), which makes the older drafts list (endpoint `/company/draft`) redundant.
- These now redundant route, use case, presenter etc got deleted 
- The drafts list is now shown ordered by creation date, descending
- There is now an html anchor `#drafts`: `/company/my_plans#drafts`. The url_index has now a corresponding method `get_my_plan_drafts_url(self)`

A bit off-topic, there is also a commit that ignores a new pytest deprecated warning.

Plan-ID: ecdb2bb1-6690-4a6e-a8a5-d08e26ac1afd


sloschert:
> I think that the 404 here is problematic since `404` means `Resource not found` but the problem with this request is not that the resource is lacking. 

please have a look at my recent commit, if the solution is fine for you.

PR #561, opened 2022-11-10
Restructure sending of confirmation emails

seppeljordan:
This PR restructures large parts of sending account confirmation emails.

The leading principle was to make the code more testable and to make the classes in use more explicit.

The idea of a `TemplateIndex` was abandoned since it leaked implementation details to higher levels of extraction. In fact the `arbeeitszeit_web` layer does not need to know about templates at all. Instead I went with the approach of a `TextRenderer` where you can request auto generated text content without having to know about how this is implemented in detail.

While working on this I decided to create a new method for the `CompanyGenerator`. This new method is responsible for creating new companies in a testing context. Instead of returning a company entity it "only" returns a company `UUID`. This makes for better decoupling from the `arbeitszeit` internals IMO.


sloschert (approved):
Great work! Certs transferred.

PR #562, opened 2022-11-11
Use cachix in github pipeline

seppeljordan:
This PR introduces the cachix service into the github pipeline. Cachix is a binary cache for the nix build tool. Using this tool all generated binaries (arbeitszeit python package and its dependencies) are uploaded to cachix services when the build pipeline succeeds. On successive build those stored binaries will be reused provided that there inputs (code or dependencies) have not changed.

This setup also allows developers to access this binary cache in the local development setups so that they don't have to rebuild anything that the CI pipeline already built in the past.

No certs needed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -276,6 +276,18 @@ Regenerate the API docs via:
 
      ./regenerate-api-docs
 
+Using a binary cache
+--------------------
+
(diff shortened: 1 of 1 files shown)


sloschert (approved):
Sounds good!

PR #570, opened 2022-11-14
Refactor login mechanisms

seppeljordan:
This PR implements a number of smaller improvements to the code quality.

1) Login to the `Session` object is now done via the user id instead of the email address of the user.
2) All registration and login related use cases now return the user id if login was successful.
3) The `FakeSession` object was simplified based on the changes to the login system.


sloschert (approved):
:+1:  CT

PR #573, opened 2022-11-14
Implement profiling dashboard via optional dependency to flask_profiler

seppeljordan:
This PR uses the package `flask_profiler` to implement a basic dashboard for performance profiling. `flask_profiler` is a optional dependency. The user would be required to explicitly install the package via the `arbeitszeitapp[profiling]` identifier.

It would be very helpful to see some performance metrics for the live instance of the application. This could serve as a first step towards that goal.

As this PR is right now, users of the `arbeitszeit_flask/development_settings.py` config should have access to the profiling dashboard right away (assuming `flask_profiler` is installed, this will happen automatically for `nix` users).

The `flask_profiler` package seems to be unmaintained. We could consider adopting it.

### DIFF
--- a/arbeitszeit/repositories.py
+++ b/arbeitszeit/repositories.py
@@ -39,6 +39,17 @@ class QueryResult(Protocol, Generic[T]):
         ...
 
 
+class PlanResult(QueryResult[Plan], Protocol):
+    def order_by_creation_date(self, ascending: bool = ...) -> PlanResult:
+        ...
+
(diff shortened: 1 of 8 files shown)


sloschert:
Does not seem to work for me. I set up development environment with `nix develop` and start flask with `python -m flask run -h localhost`. Heading over to `http://localhost:5000/profiling` gives me 404.

sloschert (approved):
With DEBUG_DETAILS on it works great.  A good idea. CT.

PR #577, opened 2022-11-15
Implement ConfirmMemberUseCase

seppeljordan:
This PR implements a proper use case for email confirmation for members.


arbeitszeit_flask/auth/routes.py
```diff
@@ -86,28 +87,16 @@ def signup_member(view: SignupMemberView):
 @auth.route("/member/confirm/<token>")
 @commit_changes
 @with_injection()
-def confirm_email_member(token, member_repository: MemberRepository):
+def confirm_email_member(
+    token, member_repository: MemberRepository, use_case: ConfirmMemberUseCase
```
sloschert, line 91: 3 details:
- member_repository can get deleted
- token misses type hint
- nested function `def redirect_invalid_request()` is now unnecessary imo

sloschert (approved):
I only had one comment, but it's approved. CT.

PR #578, opened 2022-11-15
Allow flask_profiler to be enabled in production environment

seppeljordan:
The profiler is enabled by default in development contexts (even without DEBUG_DETAILS=1) and disabled in production contexts. With this change users can enabled profiling in production contexts as well via editing the base configuration according the the flask_profiler documentation.

No certs needed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -256,7 +256,7 @@ This project uses ``flask_profiler`` to provided a very basic
 graphical user interface for response times.  More in depth profiling
 information is printed to stdout (the terminal) when detailed
 debugging is enabled. Run the following in the same terminal as you
-start the development server to enabled detailed profiling.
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -256,6 +249,19 @@ See the `flask mail documentation
...
 production backend.
 
+Profiling
+---------
+
+This project uses ``flask_profiler`` to provided a very basic
+graphical user interface for response times.  More in depth profiling
+information is printed to stdout (the terminal) when detailed
+debugging is enabled. Run the following in the same terminal as you
+start the development server to enabled detailed profiling.
```
sloschert, line 259: ...to enable...

sloschert (approved):
:+1: Only one small remark.

PR #579, opened 2022-11-18
Fix profiling issues

seppeljordan:
This PR addresses an issue with `flask-profiler`. The upstream version of flask profiler cannot deal with routes that contain a `<uuid:xyz>` url transformer. I created fork of `flask-profiler` to address this issue. This PR will make it so that pip and nix users will use this fork instead.

This is only a temporary fix. As a long term solution I suggest that we actually adopt the flask-profiler package and maintain it. A discussion on this topic should be held during plenary session of the development group.

No certs needed.


sloschert (approved):
Great.

PR #580, opened 2022-11-18
Refactor member repository

seppeljordan:
Similar to the recent refactoring of the `PlanRepository` this PR implements equivalent mechanisms for querying members. A `MemberResult` protocol was declared and methods that return multiple members were changed accordingly.


sloschert (approved):
:clap: CT :money_with_wings:

PR #575, opened 2022-11-15
Refactor `arbeitszeit.repositories.PlanRepository` and its implementations

seppeljordan:
This PR extends on the ideas of #574.  I declared and implemented more methods on the `PlanResult` interface/implementations. This made several repository methods and their tests obsolete. I deleted those.

Plan-ID: a06bef2a-629b-491e-ae76-9fa95fe5929e

PR #581, opened 2022-11-19
Refactor PurchaseRepository to use new query api

seppeljordan:
The change in this PR is pretty similar to the changes in the MemberRepository and PlanRepository conducted recently in #580 and #575 


sloschert (approved):
:+1: CT

PR #584, opened 2022-11-20
Update versions of actions if CI workflow

knit-bee:
I updated the versions of `actions/checkout`, `actions/setup-python`  and `cachix` in the CI workflow to suffice the deprecation warnings.


sloschert (approved):
:+1:

PR #585, opened 2022-11-21
Refactor `PlanCooperationRepository`

seppeljordan:
I removed a method from the `PlanCooperationRepository` protocol and moved the functionality into the `PlanResult` queryset. Also some type trickery was implemented to improve the usability of the `QueryResult` protocol and their implementations.


sloschert (approved):
Top! CT

PR #587, opened 2022-11-21
Refactor `MemberRepository`

seppeljordan:
This PR refactors the `MemberRepository` API and implementations. The new idea about the PR is that it introduces an "update" query (similar to SQL `UPDATE`). Let me know if the api design goes in the right direction.

### DIFF
--- a/tests/flask_integration/test_member_repository.py
+++ b/tests/flask_integration/test_member_repository.py
@@ -307,7 +307,6 @@ class ConfirmMemberTests(FlaskTestCase):
         expected_count = 5
         for _ in range(5):
             self.member_generator.create_member_entity()
-        print(list(self.repository.get_members()))
         assert (
             expected_count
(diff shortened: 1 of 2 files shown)


tests/flask_integration/test_member_repository.py
```diff
@@ -260,38 +272,49 @@ def setUp(self) -> None:
...
+        )
         member = self.repository.get_members().with_id(other_member_id).first()
         assert member
         assert member.confirmed_on is None
 
+    def test_that_member_confirmation_returns_the_count_of_update_members(self) -> None:
+        expected_count = 5
+        for _ in range(5):
+            self.member_generator.create_member_entity()
+        print(list(self.repository.get_members()))
```
sloschert, line 310: print statement

sloschert (approved):
there is one print statement. apart from that, it looks good! CT

PR #647, opened 2023-02-12
Refactor Api schema converter

sloschert:
This PR adds the field `total_results` to the API response when querying plans. 

There are still missing the response fields `limit` and `offset`, they will be added in a future PR.

In this context I changed a few things:
- When setting the attribute `as_list` of an json element to `True`, the element itself gets now returned as a list (also for documentation), not its children. 
- I refactored the api schema converter. I think it is more concise now.


seppeljordan (approved):
Good change :)
CT

PR #648, opened 2023-02-13
Extend query plans api presenter

sloschert:
This PR adds the fields limit and offset to the query plans api response.

To accomplish this, the presenter does not simply pass the use case response object to the marshalling decorator anymore, but creates a view model. Therefore, it has a `create_view_model`-method. (In this method are two assert statements. An alternative would be to make limit and offset in the use case request obligatory. Not sure what's preferable.)

In the presenter testing class `TestViewModelCreation` I only tested the parts that got touched in the presenter.
 
### DIFF
--- a/tests/api/presenters/test_list_active_plans_presenter.py
+++ b/tests/api/presenters/test_list_active_plans_presenter.py
@@ -19,19 +19,19 @@ class TestViewModelCreation(BaseTestCase):
         self.presenter = self.injector.get(QueryPlansApiPresenter)
         self.queried_plan_generator = self.injector.get(QueriedPlanGenerator)
 
-    def test_presenter_raises_when_no_limit_was_requested(self):
+    def test_presenter_shows_limit_as_None_when_no_limit_was_requested(self):
         use_case_response = self.queried_plan_generator.get_response(
             requested_limit=None, requested_offset=10
         )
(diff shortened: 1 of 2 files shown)


tests/api/presenters/test_list_active_plans_presenter.py
```diff
@@ -6,16 +6,75 @@
...
+            requested_limit=None, requested_offset=10
+        )
+        with self.assertRaises(AssertionError):
+            self.presenter.create_view_model(use_case_response)
+
+    def test_presenter_raises_when_no_offset_was_requested(self):
+        use_case_response = self.queried_plan_generator.get_response(
+            requested_offset=None, requested_limit=10
+        )
+        with self.assertRaises(AssertionError):
```
seppeljordan, line 33: I think that it is generally a bad idea to use `AssertianError`s a means to handle errors. Assertions in the production code should be reserved for configurations that are meant to be impossible, for example an active plan without a activation date. The reason for this is that the python interpreter has an "optimized" mode, that can be started via `python -O`. Assertions are completely ignored in this mode.

https://stackoverflow.com/questions/2055557/what-is-the-use-of-the-o-flag-for-running-python

seppeljordan (changes requested):

seppeljordan (approved):
Good work :) CT

PR #659, opened 2023-02-16
Delete left over debug print statements

seppeljordan:
This PR deletes left over print statements from previous debugging.

No certs needed.


seppeljordan:
This is now part of #660 . Therefore I am closing this.

PR #661, opened 2023-02-21
Add svg diagrams to user manual

sloschert:
This PR makes some changes to the user manual text and adds two diagrams (one for the members section, one for the companies section).

The diagrams are in svg format in order for us to be able to use gettext.

These svg diagrams were made with the help of draw.io (https://app.diagrams.net/). I added the draw.io files to the package to recreate/change them more easily in the future.

Let me know what you think about this kind of external dependency to draw.io and also the path (arbeitszeit_flask/static/diagrams).


seppeljordan:
The diagram files that you provided don't seem to be used in the template. If this is the case then why are those files in the `static` folder?

seppeljordan:
Also I have concerns regarding using drawio since it is non free software. Is there a specific reason why the raw files cannot be included as a svg file in the source code?

seppeljordan:
Generally the diagrams are a really good idea though. I like the change.

seppeljordan:
You are right regarding the free-ness of drawio. No need to use another problem if you feel comfortable with drawio.

Regarding the saving of the file in the static folder: The pupose of the static folder is to have a place for assets that need to be delivered to the client. It seems to me like the `*.drawio` files here serve a different purpose. That is why I think they should be in a different folder that calls out their purpose better.

seppeljordan (approved):
Great change. CT

PR #664, opened 2023-02-25
Pagination for QueryCompanies (1)

sloschert:
This PR starts to implement pagination for QueryCompanies. 

1) Introduce new query style to the filtering in the CompanyRepository (by name, by email).
2) Pagination in controller. Write test for controller (they were completely missing)
3) relocate QueryCompanies controller + presenter to appropriate folders.

Next steps will be pagination in use case and presenter


seppeljordan (changes requested):
Good change. I had one remark.

seppeljordan (approved):
CT

PR #669, opened 2023-03-04
Refactor database models

seppeljordan:
This PR should be reviewed **after #667 is merged**.

This PR refactors the database interface and the postgresql implementation of the interface.

The `PayoutFactorRepository` and the `PurchaseRepository` were completely removed in favor of `PayoutFactorResult`, `ConsumerPurchaseResult` and `CompanyPurchaseResult`.

Some tests were reworked to be less reliant on the database interface.

I worked more than a day on this. That's why I would appreciate the purchase of 3 PRs by the collective.

Plan-ID: b0aa02ca-667a-4028-998c-834f3186b418 (3x)


seppeljordan:
I did not get that error message when testing this change. Your comment sound like we should go ahead an merge this despite your error. Did I understand you correctly there?

PR #675, opened 2023-03-10
Pagination for QueryCompanies

sloschert:
This PR finishes pagination for QueryCompanies (Use Case, Presenter and HTML).

Regarding the presenter tests: They are largely identical to the tests for the QueryPlans presenter. In order to prevent this code duplication: Do you think it's better to test the paginator instead? Or test the paginator functionalities only in one of both presenters?

In `arbeitszeit_web.pagination.py` there is a new constant `DEFAULT_PAGE_SIZE = 15` which was necessary due to the seperation of controller and presenter in different modules.

Plan-ID: 7fbbf570-41a0-45b3-8f2d-177278c31546 (2x)

### DIFF
--- a/arbeitszeit_web/controllers/query_companies_controller.py
+++ b/arbeitszeit_web/controllers/query_companies_controller.py
@@ -54,7 +54,7 @@ class QueryCompaniesController:
                 filter_category = CompanyFilter.by_email
             else:
                 filter_category = CompanyFilter.by_name
-        offset = self._get_pagination_offset(self.request)
+        offset = self._get_pagination_offset()
         return QueryCompaniesRequestImpl(
             query=query,
             filter_category=filter_category,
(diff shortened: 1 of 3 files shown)


arbeitszeit_web/controllers/query_companies_controller.py
```diff
@@ -52,7 +54,7 @@ def import_form_data(
                 filter_category = CompanyFilter.by_email
             else:
                 filter_category = CompanyFilter.by_name
-        offset = self._get_pagination_offset(request) if request else 0
+        offset = self._get_pagination_offset(self.request)
```
seppeljordan, line 57: Since the `Request` object is now an attribute of `QueryCompaniesController` you don't need to pass `self.request` to `self._get_pagination_offset` but can access `self.request` directly in that method.

seppeljordan (approved):
Great. CT (2x)

PR #676, opened 2023-03-11
Replace out-of-date documentation on DB interface with current info

seppeljordan:
This change replaces the out-of-date docs from the development guide regarding DB interaction with info on our current approach.

Plan-ID: b0aa02ca-667a-4028-998c-834f3186b418


seppeljordan:
@sloschert All your remarks should be addressed.

PR #678, opened 2023-03-12
Companies API

sloschert:
Should not be reviewed until #675 is merged.

This PR adds the `Companies` Ressource to our API as well as the `/companies` endpoint which allows for querying companies via a GET request.

Calls to that endpoint do not need to authentificate, similar to the other existing endpoint, `plans/active`. I think authentification is, at least for now, not really needed for both endpoints. My next step would be to implement authentification for upcoming endpoints.

I'm not sure if the integration tests in `tests/api/integration/test_*.py` are too many, considering the time the tests need to run.

Plan-ID: 7fbbf570-41a0-45b3-8f2d-177278c31546 (2x)

### DIFF
--- a/arbeitszeit_flask/api/companies.py
+++ b/arbeitszeit_flask/api/companies.py
@@ -22,7 +22,7 @@ from arbeitszeit_web.api_presenters.query_companies_api_presenter import (
 namespace = Namespace("companies", "Companies related endpoints.")
 
 input_documentation = generate_input_documentation(
-    QueryCompaniesApiController().create_expected_inputs()
+    QueryCompaniesApiController.create_expected_inputs()
 )
 
 model = json_schema_to_flaskx(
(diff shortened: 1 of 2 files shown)


tests/api/controllers/test_query_companies_api_controller.py
```diff
@@ -0,0 +1,92 @@
...
+
+class ControllerTests(BaseTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.controller = self.injector.get(QueryCompaniesApiController)
+        self.request = self.injector.get(FakeRequest)
+
+    def test_that_by_default_a_request_gets_returned_which_filters_by_company_name(
+        self,
+    ):
```
seppeljordan, line 17: There are type annotations missing from this code.

seppeljordan (changes requested):

seppeljordan (approved):
Alright. Good work. CT 2x

PR #679, opened 2023-03-14
change to plan_summary parameter

sloschert:
I slightly refactored the plan_summary service method in order to be called with a plan_id (UUID) instead of a plan entity. I did this in preparation of an API route that should return the summary of a specific plan.


seppeljordan (approved):
Good job. CT

PR #680, opened 2023-03-15
Update Readme ("Getting started") and dev docs

sloschert:
I made some changes to the README file and the developer's documentation. 

The intention was to make it easier for new developers to "get started" by separating the basic dev setup ("getting started") from further development information. 

Btw: I had to delete my local `build` folder before running `make html`, to render the new dev docs correctly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -55,7 +55,7 @@ There are three user roles:
 
 .. end-introduction-do-not-delete
 
-.. start-development-setup-do-not-delete
+.. start-development-philosophy-do-not-delete
 
 Development philosophy
 =======================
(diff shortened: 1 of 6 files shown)


seppeljordan (changes requested):
Thanks for the change. I agree with cutting the content that you did.

I don't understand why this change is necessary since the structure of the document was pretty straight forward before. If a new developer wanted to start a development server they would read all the documentation under "Development setup" until `python -m flask
run -h localhost`. All the information before that was essential to get the arbeitszeitapp up and running. This seems like an intuitive way of setting up the readme. 
Also most of the stuff from the section "Further development setup" is important to newbie developers anyway, e.g. test user creation, testing, code formatting and documentation, which begs the question why these topics are not in the same section as setting up the DB.

If I had to guess I would say that this change was made to be more appealing to potential new developers and that the "Getting started" section should be as short as possible to seem not so intimidating. As I think that splitting the section about the development setup into two is compromising the structure of the readme file I don't think that this is a good trade off.

If you feel like some of the information in the readme is overwhelming then we should maybe put it in a separate document that is only rendered to by sphinx.

PR #696, opened 2023-04-10
Add section "User identification" to development guide

seppeljordan:
This change adds a new section to the development guide called "User identification". It provides info basic concepts for user account handling.


Amittai-Aviram (commented):
Sorry I did not get to this sooner.  This is helpful information.  I would have made some small stylistic changes, but I'm afraid that it is probably too late now — perhaps, in May or so, I can go through the whole document and suggest such edits.

PR #708, opened 2023-05-02
Remove nix packaging instructions for custom version of flask-restx

seppeljordan:
This change removes the custom packaging instructions for the flask-restx package. Since it is packaged inside of nixpkgs there is no need to rely on a custom version. Instead we use the default version of flask-restx that comes with nixpkgs to reduce maintenance costs.


Amittai-Aviram:
Just a note of something I wish to ask at our meeting:

Why did you all use a custom package in the first place?  (I do not really know what a custom package is anyway, but I guess I will find out.)

Amittai

Amittai Aviram
646-483-2639
***@***.***
http://www.amittai.com

PR #709, opened 2023-05-05
Fix bug due to ambigous foreign keys

### LINKED ISSUE #684: Navigating to "Purchases" lets app crash
When a user, logged in as member or a company, navigates to the "Purchases" route, the app crashes with the following error:

```
sqlalchemy.exc.AmbiguousForeignKeysError: Can't determine join between 'Join object on Join object on Join object on company_purchase(140030880489744) and anon_1(140029896719424)(140029898635248) and anon_1(140029896720096)(140029898623152) and anon_1(140029896717744)' and 'anon_1'; tables have more than one foreign key constraint relationship between them. Please specify the 'onclause' of this join explicitly.
```

Reproduced on our live instances and on master.

This error has not been detected until now because we have no integration tests for `/member/purchases` and `/company/purchases` in place.

sloschert:
fixes #684

This PR fixes the problem described in #684 which was caused by a missing onclause/ambigous foreign keys in a sql join. 

It also adds integration tests for the affected routes `/member/purchases` and `/company/purchases`.

Plan-ID: f3031122-347b-4b50-aeb4-640d81121383 (2x, Plan needs to get approved first)

### DIFF
--- a/tests/flask_integration/test_company_purchases_view.py
+++ b/tests/flask_integration/test_company_purchases_view.py
@@ -1,0 +1,62 @@
+from tests.flask_integration.flask import ViewTestCase
+
+
+class AnonymousUserTest(ViewTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.url = "/company/purchases"
+
+    def test_anonymous_user_gets_302(
+        self,
+    ):
+        response = self.client.get(self.url)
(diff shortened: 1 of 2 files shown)


seppeljordan (approved):
Excellent. Certs transferred.

PR #715, opened 2023-05-10
Move token generation logic from business layer to web layer

seppeljordan:
This change moves all the web token generation and validation from the business layer (`/arbeitszeit`) to the web layer
(`/arbeitszeit_web`). This change makes it much easier to test the business logic since it is not concerned with input validation. The input validation logic itself is also easier to test because of this change.

As part of this change two new controllers were introduced: ConfirmMemberController and ConfirmCompanyController.  These new controllers are concerned with extracting the email address encoded in the token and validating that the token is not expired yet.

For better transparency all presenter interfaces declared in the business layer (`/arbeitszeit`) were moved to `/arbeitszeit/presenters.py`.

Plan-ID: 8c5dfb71-affb-4def-b850-aafa86b2c8fd (4x)


Amittai-Aviram:
Note for a question:  What is a presenter interface?

Amittai

Amittai Aviram
646-483-2639
***@***.***
http://www.amittai.com

PR #720, opened 2023-05-16
API route `/auth/login_member`

sloschert:
This PR proceeds with our Json web API, creating an `auth/login_member` route. 

As discussed, a successful call to `login_member` will login the member in the session cookie. This design decision will require the API consumer to create a session (e.g. by using `requests.session()`) to make follow-up calls as a logged-in member. 

I had to exempt our API blueprint from CSRF protection (see: https://stackoverflow.com/questions/21509728/flask-restful-post-fails-due-csrf-protection-of-flask-wtf).

There is some kind of innovation by returning ErrorRespose objects defined in `arbeitszeit_web` in presenter and controller while defining an errorhandler on `arbeitszeit_flask`-level (`arbeitszeit_flask/api/__init__.py`). By this means there is less integration testing necessary.

There is a new decorator that can be used to easily document ocurring error responses per API method: `arbeitszeit_flask/api/response_documentation.py`

The swagger documentation looks like that now:

Bildschirmfoto vom 2023-05-16 19-15-40

Plan-ID: f3031122-347b-4b50-aeb4-640d81121383 (10x)

### DIFF
--- a/arbeitszeit_flask/api/response_documentation.py
+++ b/arbeitszeit_flask/api/response_documentation.py
@@ -12,6 +12,7 @@ class with_response_documentation:
         self._namespace = namespace
 
     def __call__(self, original_function):
+        decorated_fn = original_function
         for response in self._error_responses:
             error_schema = json_schema_to_flaskx(
                 schema=response.get_schema(), namespace=self._namespace
@@ -19,5 +20,5 @@ class with_response_documentation:
             decorator = self._namespace.response(
                 code=response.code, description=response.description, model=error_schema
             )
-            decorated_fn = decorator(original_function)
+            decorated_fn = decorator(decorated_fn)
         return decorated_fn
--- a/arbeitszeit_web/api_presenters/response_errors.py
+++ b/arbeitszeit_web/api_presenters/response_errors.py
@@ -1,9 +1,36 @@
-from abc import ABC, abstractproperty
+from typing import Protocol
 
 from arbeitszeit_web.api_presenters.interfaces import JsonDict, JsonString, JsonValue
 
 
-class ApiResponseError(ABC):
+class ApiResponseError(Protocol):
+    def __init__(self, message: str) -> None:
+        ...
+
+    @property
+    def code(self) -> int:  # must be class var
+        ...
+
+    @property
+    def description(self) -> str:  # must be class var
+        ...
+
+    @classmethod
+    def get_schema(cls) -> JsonValue:
+        ...
+
+    @property
+    def message(self) -> str:
+        ...
+
+
+class Unauthorized(Exception):
+    code = 401
+    description = "Unauthorized"
+
+    def __init__(self, message: str) -> None:
+        self.message = message
+
     @classmethod
     def get_schema(cls) -> JsonValue:
         return JsonDict(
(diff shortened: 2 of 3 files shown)


arbeitszeit_flask/api/response_documentation.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def __call__(self, original_function):
+        for response in self._error_responses:
+            error_schema = json_schema_to_flaskx(
+                schema=response.get_schema(), namespace=self._namespace
+            )
+            decorator = self._namespace.response(
+                code=response.code, description=response.description, model=error_schema
+            )
+            decorated_fn = decorator(original_function)
```
seppeljordan, line 22: This doesn't look right. I seems like only ever the last decorator constructed is actually applied to the original function.

arbeitszeit_web/api_controllers/login_member_api_controller.py
```diff
@@ -0,0 +1,32 @@
...
+            ),
+            ExpectedInput(
+                name="password",
+                type=str,
+                description="Password.",
+                default=None,
+            ),
+        ]
+
+    def create_request(self, request: Request) -> LogInMemberUseCase.Request:
```
seppeljordan, line 27: Consider injecting the `Request` object instead of passing it into the controller.

arbeitszeit_web/api_presenters/response_errors.py
```diff
@@ -0,0 +1,28 @@
+from abc import ABC, abstractproperty
+
+from arbeitszeit_web.api_presenters.interfaces import JsonDict, JsonString, JsonValue
+
+
+class ApiResponseError(ABC):
```
seppeljordan, line 6: Do you think that this could be a `Protocol` instead? This would allow us to implement the `ApiResponseError` without inheriting from this `ABC`.

tests/api/presenters/test_login_member_presenter.py
```diff
@@ -0,0 +1,122 @@
...
+from tests.session import FakeSession
+
+
+class TestViewModelCreation(BaseTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.presenter = self.injector.get(LoginMemberApiPresenter)
+        self.session = self.injector.get(FakeSession)
+
+    def test_correct_error_raises_if_wrong_mail_adress_was_given(self) -> None:
```
seppeljordan, line 21: I would prefer if you specified the error type already in the test name. `correct_error` is kind of vague.

arbeitszeit_web/api_controllers/login_member_api_controller.py
```diff
@@ -0,0 +1,32 @@
...
+        email = request.get_form("email")
+        password = request.get_form("password")
+        if not email or not password:
+            raise Unauthorized(message="Email or password missing.")
```
seppeljordan, line 31: I think that we should not respond `Unauthorized` here but instead `BadRequest` of even `OK` since the problem with the incoming request is that it is missing data and not that it is "unauthorized".

seppeljordan (changes requested):
I like where this is heading but I had some remarks regarding naming and HTTP.

arbeitszeit_flask/api/response_documentation.py
seppeljordan: You are right, it works. The reason why this works is that `flask-restx` violates the convention on how decorators should work.

Usually a decorator accepts a function as an argument and returns a new function while leaving the old function unchanged. Although `Namespace.response` from `flask-restx` does return a new function it also modifies the original function.

See here for the implementation: https://github.com/python-restx/flask-restx/blob/master/flask_restx/namespace.py#L129

Since this is code outside of our project it is probably okay to leave it as is. I would still prefer

```python
decorated_fn = original_function # <---
for response in self._error_responses:
    ...
    decorated_fn = decorator(decorated_fn)  # <-- decorated_fn instead of original_function
return decorated_fn
```

seppeljordan (approved):
I had one minor remark.
Really great work.
CT 10x

seppeljordan (approved):

PR #721, opened 2023-05-18
New query parameters for `/company/transfer_to_company`

sloschert:
This PR proposes to add two optional query parameters (amount and type_of_payment) to the `/company/transfer_to_company` route. This would allow us to generate easily pay links (that we could use e.g. here on Github to simplify the pay process).

Plan-ID: f3031122-347b-4b50-aeb4-640d81121383 (2x)

# Commits:

### Optional param amount in pay consumer product url
In url_index the param `amount` was non-optional. This was a unnecessary constraint. The param is now opti
onal.

### PayMeansOfProductionForm: rename form field
Rename from "category_field" to "type of payment_field"

### two more params for pay_means_of_production_url
add `amount` and `type_of_payment` as params for url creation in url_index

### possible type_of_payments values to lowercase
"fixed" and "liquid" instead of "Fixed" and "Liquid"

### DIFF
--- a/arbeitszeit_flask/views/pay_means_of_production.py
+++ b/arbeitszeit_flask/views/pay_means_of_production.py
@@ -1,8 +1,7 @@
 from dataclasses import dataclass
-from typing import Optional
 
 from flask import Response as FlaskResponse
-from flask import redirect, request
+from flask import redirect
 
 from arbeitszeit.use_cases.pay_means_of_production import PayMeansOfProduction
 from arbeitszeit_flask.forms import PayMeansOfProductionForm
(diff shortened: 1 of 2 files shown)


arbeitszeit_flask/views/pay_means_of_production.py
```diff
@@ -23,8 +23,14 @@ class PayMeansOfProductionView:
 
     def respond_to_get(self, form: PayMeansOfProductionForm) -> Response:
         plan_id: Optional[str] = request.args.get("plan_id")
+        amount: Optional[str] = request.args.get("amount")
+        type_of_payment: Optional[str] = request.args.get("type_of_payment")
         if plan_id:
             form.plan_id_field().set_value(plan_id)
+        if amount:
```
seppeljordan, line 30: This looks wrong to me. Shouldn't the form already be constructed with the correct fields in it?

seppeljordan (changes requested):
Great PR. I had some issues with one part of the view code though.

seppeljordan (approved):
CT (2x)

PR #722, opened 2023-05-20
Refactor integration tests

sloschert:
This is a proposel to refactor our flask integration tests. Motivation is that many integration tests (tests for expected status codes) are missing and that there should be a simpler way to write these tests. 

(We could discuss though if it is really necessary to check with integration tests the returned status code of every request method of every route with every possible user) 

### add python package "parameterized" to requirements

This commit would add the package ["parameterized"](https://github.com/wolever/parameterized) to our code for parameterized tests. Reason is that pytest's `mark.parameterized` decorator does not work with unittest's classes and the functionality of the built-in unittest is not sufficient.

### New method on ViewTestCase

The new method `ViewTestCase().assert_response_has_expected_code()` should make it a bit easier to write integration tests.

### Refactoring of some integration tests

Apply the new "parameterized" package and the new method `self.assert_response_has_expected_code` to refactor some integration tests. Before refactoring more tests I wanted to hear your feedback.

### `tests.markers`: fix problem with type-checking

At least in my setting (vs code + pylance) autocompletion in ViewTestCase was not working. In this commit I propose a fix.

Plan-ID: f3031122-347b-4b50-aeb4-640d81121383 (4x)

### DIFF
--- a/nix/pythonPackages/arbeitszeitapp.nix
+++ b/nix/pythonPackages/arbeitszeitapp.nix
@@ -11,7 +11,7 @@ buildPythonPackage {
   outputs = [ "out" "doc" ];
   postPhases = [ "buildDocsPhase" ];
   format = "pyproject";
-  buildInputs = [ pytestCheckHook sphinx ];
+  buildInputs = [ pytestCheckHook sphinx parameterized ];
   propagatedBuildInputs = [
     email_validator
     flask
@@ -25,7 +25,6 @@ buildPythonPackage {
     is_safe_url
     matplotlib
     typing-extensions
-    parameterized
   ];
   buildDocsPhase = ''
     mkdir -p $doc/share/doc/arbeitszeitapp
--- a/tests/markers.py
+++ b/tests/markers.py
@@ -1,12 +1,13 @@
 import enum
 import os
 from typing import Set, TypeVar
+from unittest import TestCase
 
 _MARKERS: Set[str] = {
     marker.strip() for marker in os.getenv("DISABLED_TESTS", "").split(",")
 }
 
-T = TypeVar("T")
+T = TypeVar("T", bound=TestCase)
 
 
 class Category(enum.Enum):


nix/pythonPackages/arbeitszeitapp.nix
```diff
@@ -25,6 +25,7 @@ buildPythonPackage {
     is_safe_url
     matplotlib
     typing-extensions
+    parameterized
```
seppeljordan, line 28: This should go into `buildInputs` instead.

tests/flask_integration/flask.py
```diff
@@ -152,3 +159,29 @@ def _confirm_company(
             follow_redirects=True,
         )
         assert response.status_code < 400
+
+    def assert_response_has_expected_code(
```
seppeljordan, line 163: I think that this function is problematic. It takes way to many arguments for what it does and is basically just a wrapper around `self.client.get` and `self.client.post`. Further the `login` argument changes the state of the client in a unexpected way. Before the method call the test client might not be authenticated but after calling this method the client might be authenticated. I find this behavior surprising. Looking at where you used this method it also did not make the code more concise.

```
self.login_member()
assert self.client.get(url).status_code == expected_status_code
```
is more concise than

```
self.assert_response_has_expected_code(
    url=expected_url,
    method='get',
    login=LoginUser.member,
    expected_code=expected_status_code,
)
```

tests/flask_integration/test_accountant_dashboard_view.py
```diff
@@ -1,30 +1,29 @@
...
-        self.login_member()
-
-    def test_does_not_return_200_response(self) -> None:
-        response = self.client.get("/accountant/dashboard")
-        self.assertNotEqual(response.status_code, 200)
+        self.url = "/accountant/dashboard"
+
+    @parameterized.expand(
+        [
+            (LogInUser.accountant, 200),
```
seppeljordan, line 15: One could argue that we don't need to exercise all 4 cases for this route since this behavior is not even part of the view code but instead of the authentication logic. I would IMO suffice to test this only for an accountant and then again for a non-accountant. This would already demonstrate that we use the proper authentication decorator in our flask code.

Those two remaining test cases should then also be separated into individual test cases since the name of the current test is really not descriptive.

tests/markers.py
```diff
@@ -1,12 +1,13 @@
...
 import os
-from typing import Set, Type
-from unittest import TestCase
+from typing import Set, TypeVar
 
 _MARKERS: Set[str] = {
     marker.strip() for marker in os.getenv("DISABLED_TESTS", "").split(",")
 }
 
+T = TypeVar("T")
```
seppeljordan, line 9: I think that this type variable should be bound to `TestCase` since it is used in a context where we would only ever want descendants of `TestCase` to be passed in.

seppeljordan (changes requested):
I think that this PR has some stylistic problems but goes into the right direction. Let me know if you find my criticism unwarranted.

tests/flask_integration/flask.py
seppeljordan: I agree with the arguments you made and think that this change should go forward.

tests/flask_integration/test_accountant_dashboard_view.py
seppeljordan: With the context from [above](https://github.com/arbeitszeit/arbeitszeitapp/pull/722#discussion_r1199751511) this comment is obsolete.

seppeljordan (approved):
Great change. CT 4x

PR #729, opened 2023-05-28
Update develoment dependencies

seppeljordan:
This PR implements an update script that allows developers to update the development dependencies in one go. One can run `python -m arbeitszeit_development.update_dependencies` from the root of this project to update the nix flake inputs, flask-restx and bulma.

Plan-ID: 8c5dfb71-affb-4def-b850-aafa86b2c8fd (2x)


seppeljordan:
@sloschert Thank you for catching those errors. I think I addressed all your remarks.

PR #730, opened 2023-05-29
API documentation: "Required" properties for input parameters and response models

sloschert:
The commits in this PR are mainly concerned with adding "required" properties to the API inputs and API response models in compliance with OpenAPI 2.0 ("Swagger") requirements. 

- Input parameters: With this PR we can define for our OpenAPI 2.0 documentation if input parameters are required or optional.
image

- Response models: As far as I have read, OpenAPi 2.0 does not allow to mark properties of a response model to be optional. The JSON schema `AnyOf[null, "something"]` is not supported. So I suppose the only way to mark properties to be optional is to mark all properties as "required" unless they are optional.  
image
This feature will get relevant in upcoming PRs, when we have view models with fields that may be None (e.g. `PlanSummary.cooperation`). These fields won't be send to the users at all, if they are None (this is accomplished by flask-restx marshalling with `skip_none=True`).

Lastly, in the schema converter a better model comparison was needed to compare equality of nested instances of a class. Instead of the basic statement `if self.namespace.models[schema_name] == model: ...`, I propose to make use of the `DeepDiff` library: `if not DeepDiff(self.namespace.models[model.name], model): ...` 

[CT-Link](https://app.arbeitszeitrechnung.org/company/transfer_to_company?plan_id=f3031122-347b-4b50-aeb4-640d81121383&amount=5) (5x)

### DIFF
--- a/arbeitszeit_flask/api/response_handling.py
+++ b/arbeitszeit_flask/api/response_handling.py
@@ -17,8 +17,7 @@ class error_response_handling:
         self._register_error_handlers()
 
     def __call__(self, original_function):
-        decorated_fn = self._add_documention_of_error_responses(original_function)
-        return decorated_fn
+        return self._add_documention_of_error_responses(original_function)
 
     def _register_error_handlers(self) -> None:
         for error in self._error_responses:


arbeitszeit_flask/api/response_handling.py
```diff
@@ -4,14 +4,27 @@
...
+class error_response_handling:
     def __init__(
         self, error_responses: list[type[ApiResponseError]], namespace: Namespace
     ) -> None:
         self._error_responses = error_responses
         self._namespace = namespace
+        self._register_error_handlers()
 
     def __call__(self, original_function):
+        decorated_fn = self._add_documention_of_error_responses(original_function)
```
seppeljordan, line 20: `return self._add_documention_of_error_responses(original_function)`

seppeljordan (approved):
Good work. I had one minor comment which be considered a suggestion rather then an issue.
CT x5

PR #743, opened 2023-06-21
Update API docs

seppeljordan:
This change updates the API docs via the `./regenerate-api-docs` command. A minor adjustment was done to that script were previous versions of the api documentation are deleted before regenerating them from the source code.


Amittai-Aviram (commented):
We would still preserve the history of doc changes in Git, right?  This looks like a good change.  Why did you delete some *.rst files?

Amittai-Aviram (commented):
BTW, why are we making this change?  Does it make the build process go faster?  Without this change, doesn't the build process simply write over the existing files?

seppeljordan:
@Amittai-Aviram 
This change is IMO necessary because sphinx does not automatically generate the api docs. At least I was not able to set it up in such a way.  Currently we are using `sphinx-apidoc` to create the module documentation from our code/docstrings. I would be happy to configure sphinx in a way that these kind of maintenance commits would be obsolete. Until then we have to run the `regenerate-api-docs` program from time to time to keep the online documentation up to date.

seppeljordan:
> We would still preserve the history of doc changes in Git, right? This looks like a good change. Why did you delete some *.rst files?

Those `.rst` files that I deleted were obsolete, meaning that they refer to modules that don't exist anymore. This might also be the reason for the failure messages that you experienced why you tried to build the docs locally.

Amittai-Aviram (approved):

PR #750, opened 2023-06-26
Aviram/update doc

Amittai-Aviram:
This set of changes to the documentation is not quite complete, because I am still trying to fix issues related to development on a Mac — particularly for files residing in iCloud.  But most of the changes should apply in any case.  I focused on making the tips for using Nix more explicit.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,4 +1,4 @@
-|Generic badge|
+Generic badge|
 
 .. |Generic badge| image:: https://github.com/arbeitszeit/arbeitszeitapp/actions/workflows/python-app.yml/badge.svg
    :target: https://github.com/arbeitszeit/arbeitszeitapp/actions/workflows/python-app.yml
@@ -14,7 +14,7 @@ Description
 ------------
 
 Arbeitszeitapp is a plattform to exchange services and products on the
-basis of working time. It is designed as a web app to be self-hosted by communities
+basis of working time. It is designed as a Web app to be self-hosted by communities
 or organizations. A test instance is running on
 https://demo-app.arbeitszeitrechnung.org/.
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -1,4 +1,4 @@
-|Generic badge|
+Generic badge|
```
seppeljordan, line 1: I looks like you accidentally deleted the first character in this line.

README.rst
```diff
@@ -133,50 +145,82 @@ environment variables:
      export ARBEITSZEIT_APP_SERVER_NAME=localhost:5000
      export ARBEITSZEITAPP_TEST_DB="postgresql://postgres@localhost:5432/<name of test database>"
 
-After configuring the database connection you need to run the database
+If you are using Nix instead of ``venv``, you can safely omit four of
```
seppeljordan, line 148: I am not so sure about this statement. In my experience is necessary to either set the same environment variables as shown above when using nix or use a custom settings file á la `custom_settings.py`.

seppeljordan (commented):
Thank you for the changes to the documentation. I think that there is one problematic passage regarding environment variables but maybe I am missing something?

README.rst
Amittai-Aviram: Thanks, Sebastian!  You are right about this and the other issue regarding environment variables.

README.rst
Amittai-Aviram: Yes!  You are right!  Sorry --- I did not fully understand how the settings files worked.  Hopefully, my corrected version is reasonably accurate.  Please let me know -- and thank you!

seppeljordan (approved):
Great change :) Thanks. I think the README is much better now.

README.rst
```diff
@@ -270,17 +354,17 @@ We use `Flask-Babel <https://python-babel.github.io/flask-babel/>`_ for translat
     python setup.py compile_catalog
 
 
-Email configuration
+E-mail Configuration
```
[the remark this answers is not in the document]
Amittai-Aviram: Thank you so much for catching that!  I think I have fixed it now.

PR #754, opened 2023-07-03
Relocating API code

sloschert:
This PR consists of 2 commits, it's only relocating some API code. No certs needed.

**Move api_controller and api_presenter into own folder**

Now there is a new folder `arbeitszeit_web/api` for api-specific code on arbeitszeit_web-level (similar to the structure in `arbeitszeit_flask/api`). Two existing folders, api_controller and api_presenter, have been moved there.

**Move the api response errors module one level up.**

The module `response_errors.py` has been in the folder of arbeitszeit_web/api/presenters. But it's code is used also by controllers and flask_restx views. That's why it has been moved one level up to `arbeitszeit_web/api`.


seppeljordan (approved):
Good change. Do you think that we should put the web presenter and controllers into their own `www` or `web` folder?

PR #756, opened 2023-07-06
Authentication check on one API endpoint

sloschert:
I have implemented a first authentication check on one of our API endpoints. If you agree I will proceed with other endpoints.

Plan: f3031122-347b-4b50-aeb4-640d81121383 (2x)

**check user is authenticated when listing plans**

A check has been implemented on the API endpoint /plans/active. It asserts that the user is authenticated and raises a 401 if they aren't. Also a short description on our API authentication process was added to the API docs.

**Move Authenticatrs to arbeitszeit_web.www-folder** 

The MemberAuthenticator and CompanyAuthenticator seem to be www-specific, that's why they were moved to the `www` directory.


seppeljordan (changes requested):
The implementation looks good to me but I would like to see some stylistic improvements.

arbeitszeit_web/api/authentication.py
```diff
@@ -0,0 +1,17 @@
...
+from arbeitszeit_web.api.response_errors import Unauthorized
+from arbeitszeit_web.session import Session
+
+
+@dataclass
+class Authenticator:
+    session: Session
+
+    @property
+    def authenticate_request(self) -> None:
```
seppeljordan, line 12: This is not a property but a method. I think that the `@property` decorator obfuscates the fact that `authenticate_request` has a side effect.

seppeljordan (approved):
CT 2x. I really would like to see the one change I commented before merging this.

PR #762, opened 2023-07-09
API: Authenticate requests of 2 API endpoints

sloschert:
Authenticate requests of `/plans/{plan_id}` and `/companies`. Only logged in members or companies are now able to use these endpoints.

Plan: f3031122-347b-4b50-aeb4-640d81121383

### DIFF
--- a/arbeitszeit_flask/api/plans.py
+++ b/arbeitszeit_flask/api/plans.py
@@ -59,8 +59,9 @@ plan_get_model = SchemaConverter(namespace).json_schema_to_flaskx(
 class Plan(Resource):
     @namespace.marshal_with(plan_get_model, skip_none=True)
     @error_response_handling(
-        error_responses=[BadRequest, NotFound], namespace=namespace
+        error_responses=[BadRequest, NotFound, Unauthorized], namespace=namespace
     )
+    @authentication_check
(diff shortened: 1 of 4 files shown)


arbeitszeit_flask/api/plans.py
```diff
@@ -59,8 +59,9 @@ def get(
 class Plan(Resource):
     @namespace.marshal_with(plan_get_model, skip_none=True)
     @error_response_handling(
-        error_responses=[BadRequest, NotFound], namespace=namespace
+        error_responses=[BadRequest, NotFound, Unauthorized], namespace=namespace
```
Amittai-Aviram, line 62: How is this list used, and why is it in this order?  It seems odd to me, because clients would have to refer to the error_responses element by index (or iterate over the list?), whereas I would expect the client to be clearer if it just specified, e.g., "Unauthorized."  BTW, what is the type of the elements in this list?

Amittai-Aviram (commented):
Looks great!  Just one question — more for my own information.

Amittai-Aviram (commented):
Toll!  Danke für Deine Erklärung.

Amittai-Aviram (approved):

PR #767, opened 2023-07-14
Specify preferred url scheme in base config

sloschert:
This small PR proposes two changes. No work registration needed.

**Add preferred url schema https to config** 

The links that we send out via mails start with `http`. This config entry changes it to `https`

**delete dead code: method render_to_html**


seppeljordan (approved):
great 👍🏼

PR #770, opened 2023-07-15
Implement GetAccountantAccountDetails view

seppeljordan:
The implemented controller, presenter and flask route allow accountants to review their on account info, which is at the moment only their own email address and their user id.

Plan-ID: a6a663c1-c041-4085-a627-c4f3395928bf


seppeljordan:
@sloschert I think that should go into a separate PR since it has nothing to do with the subject of this change. I agree that a dedicated authenticator should be implemented.

Issue #654, opened 2023-02-15
Sign in not possible on wrong capitalisation of mail address


philipbroistedt:

Amittai-Aviram:
Addressed in Pull Request #776:  https://github.com/arbeitszeit/arbeitszeitapp/pull/776#issue-1808820195

PR #776, opened 2023-07-18
Aviram/normalize mail addresses

Amittai-Aviram:
These changes are meant to implement and enforce the following semantics for e-mail addresses:  (1) We preserve the original e-mail address as first submitted in the act of creation of any entity that has an associated e-mail address. (2) When we test for an already-existing record that has the same e-mail address, we search in a case insensitive manner.  This second point applies both (a) to rejection of potential duplicate entries and (b) to acceptance of imperfect queries intended to get a record by e-mail address.  (Addresses Issue #654.)


Amittai-Aviram:
Thank you so much, Jordi!  I have run automatic formatting — but I do not understand the current test failures. :-(

PR #779, opened 2023-07-18
Remove CannotSendMail and PlanIsInactive exception types

seppeljordan:
The two exceptions that were removed in this change were largly obsolete. With some minor adjustment to the PayConsumerPurchase use case the code functions exactly as it did before.

*No registration in the app required*


arbeitszeit_flask/mail_service.py
```diff
@@ -50,7 +49,4 @@ def send_message(
         sender: str,
     ) -> None:
         msg = Message(subject=subject, recipients=recipients, html=html, sender=sender)
-        try:
-            mail.send(msg)
-        except Exception:
-            raise CannotSendEmail
+        mail.send(msg)
```
Amittai-Aviram, line 52: Does sending a mail message never raise an exception from somewhere in the process?

Amittai-Aviram (commented):
This looks good, but I would like to understand these changes a bit better.

Amittai-Aviram (commented):
Also -- should there be any tests for this change?

arbeitszeit_flask/mail_service.py
Amittai-Aviram: OK, good point.

PR #786, opened 2023-07-23
Fix bug in AccountResult.joined_with_balance

seppeljordan:
The bug that was fixed would lead to the wrong account balance being displayed. A regression test was provided.

Plan-ID: a6a663c1-c041-4085-a627-c4f3395928bf


Amittai-Aviram (approved):
LGTM.

PR #819, opened 2023-09-19
Changes to type annotations + imports in repositories.py

seppeljordan:
# Reorganize imports in arbeitszeit.repositories

Before this commit the imports in `arbeitszeit.repositories` were a
little unorganized and inconsitent, especially the the imports from
`arbeitszeit.records`. This change unifies all imports from that
module such that only the module `arbeitszeit.records` is brought into
scope and the individual type definitions are accessed via the dot
operator.

Before:
```python
def create_member(...) -> Member:
    ...
```

Now:
```python
def create_member(...) -> records.Member:
    ...
```

# Simplify type signatures on DatabaseGateway et al

Before this change we made use of a crutch when it comes to type
signatures. This change uses the new `Self` typing feature of python
in all `*Result` types for database interfacing.  The type signatures
should be easier to understand and reproduce with this change. This
simplification does come with a cost though.  After this change we
generally disallow that a specific subtype of `QueryResult` may yield
another subtype of QueryResult that is not a subtype of itself. The
author considers this trade off worthwhile since he believes that
something in the design process must have gone horribly wrong if the
need for such a monstrosity ever arises.

_Registration not needed_

### DIFF
--- a/arbeitszeit/repositories.py
+++ b/arbeitszeit/repositories.py
@@ -10,17 +10,16 @@ from typing_extensions import Self
 from arbeitszeit import records
 
 T = TypeVar("T", covariant=True)
-QueryResultT = TypeVar("QueryResultT", bound="QueryResult")
 
 
 class QueryResult(Protocol, Generic[T]):
     def __iter__(self) -> Iterator[T]:
         ...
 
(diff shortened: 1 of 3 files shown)


Amittai-Aviram (approved):
I like these changes very much!

PR #836, opened 2023-09-25
Implement EmailChangeConfirmation presenter

seppeljordan:
This change implements a rudimentary presenter for email change notifications. This functionality is not yet integrated into the application since multiple pieces of this remain to be implemented.

The test implementation is not ideal since we interrogate the email content via its expected implementation. Therefore the tests for email content are expected to be quite fragile/break often.

Plan-ID: a6a663c1-c041-4085-a627-c4f3395928bf (2x)

### DIFF
--- a/arbeitszeit_web/email/email_sender.py
+++ b/arbeitszeit_web/email/email_sender.py
@@ -4,6 +4,7 @@ from arbeitszeit import email_notifications as interface
 
 from .accountant_invitation_presenter import AccountantInvitationEmailPresenter
 from .cooperation_request_email_presenter import CooperationRequestEmailPresenter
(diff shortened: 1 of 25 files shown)


arbeitszeit_web/email/email_sender.py
```diff
@@ -44,4 +46,6 @@ def send_email(self, message: interface.Message) -> None:
         elif isinstance(message, interface.CooperationRequestEmail):
```
Amittai-Aviram, line 46: As we discussed, I am slightly uneasy about this design pattern, but I like your proposal of (eventually) replacing this logic with a dictionary lookup.

PR #840, opened 2023-10-01
Coop summary: show "end coop" buttons

sloschert:
After this commit, the "end cooperation" buttons in the associated plan list of the coop summary page are not only shown when the requester of the page is the coordinator of the cooperation, but also if it is the planner of the associated plan. Both should have the right to retrieve the plan from the cooperation. Also, the presenter tests have been refactored.

Plan: fe6ed75b-5ad9-4821-a399-7ca8303e2fea (2x)

---

image


Amittai-Aviram (approved):
LGTM!  (See question.)

Issue #317, opened 2022-03-11
Transfer coop coordination to other company


sloschert:
Right now it is not possible for a cooperation coordinator to transfer the "coordinator role" to another company. 

We still have to discuss how this transferal of coordination should work in detail. Perhaps similar to "invite worker to company", i.e. by sending a message to a company which can then accept or decline.

sloschert:
Not quite sure. Perhaps it was a mistake. Later we still should create RejectCoordinationTransfer and CancelCoordinationTransfer use cases, but we can create new issues for that. So I propose to close it.

PR #849, opened 2023-10-22
Introduce RequestCoordinationTransferUseCase

sloschert:
This commit introduces a new use case, RequestCoordinationTransferUseCase (see issue #317).  It allows current coordinators of a cooperation to request the transferal of the coordination role to another company. A mail will be sent out. Together with this use case comes a new `record`, called CoordinationTransferRequest.  Note that this use case fails if the requesting company (i.e. coordination tenure) already has requested a transfer to another company. This behaviour makes probably sense, because we expect companies to agree "in real life" which company will take over an existing coordination role - there will be no need for a bunch of requests to be sent out. Note also that the newly created CoordinationTenureGenerator creates the coordination tenure via direct access to the database_gateway, because we have not yet all necessary use cases in place to do this via use cases. The missing counterpart of this use case, called probably AcceptCoordinationTransferUseCase, is still to be written. Also the www implementation is still missing.
Plan: fe6ed75b-5ad9-4821-a399-7ca8303e2fea (6x)

### DIFF
--- a/arbeitszeit/records.py
+++ b/arbeitszeit/records.py
@@ -137,7 +137,6 @@ class CoordinationTransferRequest:
     request_date: datetime
     transfer_date: Optional[datetime]
 
-    @property
     def is_open(self) -> bool:
         return self.transfer_date is None
 
--- a/arbeitszeit/use_cases/request_coordination_transfer.py
+++ b/arbeitszeit/use_cases/request_coordination_transfer.py
@@ -115,4 +115,4 @@ class RequestCoordinationTransferUseCase:
             .requested_by(requesting_tenure.id)
             .that_are_open()
         )
-        return True if len(transfer_requests) > 0 else False
+        return len(transfer_requests) > 0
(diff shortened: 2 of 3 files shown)


arbeitszeit/records.py
```diff
@@ -129,6 +129,19 @@ class CoordinationTenure:
...
 
+@dataclass
+class CoordinationTransferRequest:
+    id: UUID
+    requesting_coordination_tenure: UUID
+    candidate: UUID
+    request_date: datetime
+    transfer_date: Optional[datetime]
+
+    @property
```
Amittai-Aviram, line 140: I just read some documentation on the \@property decorator — but I am still wondering why you are using this decorator for is_open.  Is that so that you can write transferRequest.is_open instead of transferRequest.is_open()?

Amittai-Aviram (commented):
Just a question for now.  Will try to continue reviewing.

arbeitszeit/records.py
Amittai-Aviram: In terns of clean code, I think it would be better to use the function directly rather than to hide it as a property.  In efficiency terms, the implementation of @property also adds tiny but unnecessary overhead.

Amittai-Aviram (approved):
Looks great!  I just left a couple of little comments for you to address as you wish.

PR #878, opened 2023-11-20
Have only member blueprint use `/member/` url prefix

seppeljordan:
Before this change there were a couple of routes in `auth` blueprint that were using `/member/` in there urls as a prefix. This commit changes those routes to not use the prefix in question anymore. This allows us to use `/member` as a fixed url prefix for the `member` blueprint.

Plan-ID: b412d2bb-074a-489b-86f6-90b87077c17a


Amittai-Aviram (approved):
Is the goal of this change tidiness, because it shortens the route specification (but not the actual URL) for member-blueprint requests, or is it some gained efficiency from treating "/member" as a URL prefix?

PR #897, opened 2023-12-09
Unify 404 error view and create 403 and 409 error view

sloschert:
**Use same template to render 404 for all users** 

Since our base template can figure out now, which user is currently logged in, there is no need to have different 404 templates per user.

**Create templates for 403 and 409 http errors**

_Plan: fd00d0bb-0ea3-4338-b097-dfa605278f1c_

### DIFF
--- a/arbeitszeit_flask/templates/user/403.html
+++ b/arbeitszeit_flask/templates/user/403.html
@@ -1,0 +1,9 @@
+{% extends "base.html" %}
+
+{% block content %}
+
+<div class="section has-text-centered">
+    HTTP 403 FORBIDDEN
+</div>
+
+{% endblock %}
--- a/arbeitszeit_flask/templates/user/409.html
+++ b/arbeitszeit_flask/templates/user/409.html
@@ -1,0 +1,9 @@
+{% extends "base.html" %}
+
+{% block content %}
+
+<div class="section has-text-centered">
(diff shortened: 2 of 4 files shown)


arbeitszeit_flask/templates/user/403.html
```diff
@@ -3,7 +3,7 @@
 {% block content %}
 
 <div class="section has-text-centered">
-    HTTP 404 NOT FOUND
+    HTTP 403 NOT FOUND
```
Amittai-Aviram, line 6: s/NOT FOUND/FORBIDDEN/  :-)

arbeitszeit_flask/templates/user/409.html
```diff
@@ -0,0 +1,9 @@
+{% extends "base.html" %}
+
+{% block content %}
+
+<div class="section has-text-centered">
+    HTTP 409 NOT FOUND
```
Amittai-Aviram, line 6: s/NOT FOUND/CONFLICT/  — Do we need to support this status code?  In what scenario?

Amittai-Aviram (commented):
It looks as if there might be something wrong with the Flask templates invoked by redner_template — ?  I do not understand why the wrong strings appear next to the status codes in the generated HTML files.  I know — it's not that important — but, still, I think it's incorrect — unless I am misunderstanding something.

PR #906, opened 2023-12-30
Introduce class based view interface

seppeljordan:
This change introduces a new way to define flask routes based on classes. The basic idea is that you can leverage dependency injection with views by defining a dataclass or a `__init__` method for your view that accepts all the necessary dependencies and do any setup inside that classes constructor. Then additionally you would have different python methods defined on that class to handle the different HTTP methds, e.g. "GET" or "POST".

Also all class based request handlers that existed before this change were adapted to fit the new technique introduced.

Plan-ID: b412d2bb-074a-489b-86f6-90b87077c17a (4x)


sloschert (approved):
A good change! 4xCR

PR #918, opened 2024-01-14
Use unsafe, unsecure password hasher in tests

seppeljordan:
The time it takes to execute all tests increased since commit 778769e135b26c1fdc2a7f1f5de4cd32dde2dc93 significantly. The reason for this increase is the password hashing algorithm that was introduces. The more secure password hashing alogrithm that we used is very slow. This slowness is one of the reasons why it is secure. Security is not required when running the tests. This is why after this commit we can configure differnt PasswordHasher classes to be used with the app. In the tests we now load a trivial password hasher class whereas the default configuration of the arbeitszeitapp makes use of the same class as we provided before this change (e.g. the slow secure one).

Plan-ID: b412d2bb-074a-489b-86f6-90b87077c17a (1x)

### DIFF
--- a/arbeitszeit/injector.py
+++ b/arbeitszeit/injector.py
@@ -23,7 +23,15 @@ TypeT = TypeVar("TypeT", bound=Type)
 
 class Injector:
     def __init__(self, modules: List[Module]) -> None:
+        class _RecursionModule(Module):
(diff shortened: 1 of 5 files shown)


sloschert (approved):
👍 1xCR

PR #920, opened 2024-01-21
Minor change in /member/consumptions route

seppeljordan:
Before this change we decided on the status code to reply with in the flask route itself. After this change this decision is made in the controller associated with this route.

*No registration of consumption required*


sloschert (commented):
Usually we would save the  `query_private_consumptions.py` file in `arbeitszeit_flask.views` - is it intended that it is saved in arbeitszeit_flask.member ?

sloschert (changes requested):
I have one question, apart from that it looks good to me.

sloschert:
@seppeljordan i think it would be more consistent, that is why i have a preference to move it, but i would not oppose leaving it as is.

sloschert (approved):

PR #925, opened 2024-01-22
Refactor all member routes into class based views

seppeljordan:
Plan-ID: b412d2bb-074a-489b-86f6-90b87077c17a


sloschert (approved):
:+1: 1xCR

PR #928, opened 2024-01-25
HTML for RviewRegisteredConsumptions

sloschert:
This commit adds the html layer for the company's review consumptions feature. This view allows companies to see a list of their "sales", with the buyer's non-anonymized data. If the consumer is a member, only the user's name is shown - in the long run this might not be detailed enough and we probably will have to identify the member with a link to their profile (currently there are no member profile pages that can be accessed by other users) or the member's ID number.

Builds on #926 and #927

Plan: fd00d0bb-0ea3-4338-b097-dfa605278f1c

### DIFF
--- a/arbeitszeit/use_cases/review_registered_consumptions.py
+++ b/arbeitszeit/use_cases/review_registered_consumptions.py
@@ -18,7 +18,7 @@ class RegisteredConsumption:
 
 
 @dataclass
-class RewiewRegisteredConsumptionsUseCase:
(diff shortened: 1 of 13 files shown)


sloschert:
> LGTM.
> 
> Since it took me so long to review this PR the plan that was referenced in the description expired. Let me know another Plan-ID so that I can register the consumption of your service.

Thanks. Plan-ID: b05bef9c-4dbf-4cd1-9e99-59219dfa9c4b

PR #933, opened 2024-01-31
Add aarch64-darwin as a supported systemto flake.nix

rhklee:


rhklee:
> Great :)
> 
> I have no way of validating the correctness of this but I trust that it works on your system. Feel free to merge this :)

Yes! It's now working on my machine and I can run the project locally with nix as well as the tests (well the non-database tests, running all the tests seems like it would take an extremely long time).

petsch0:
I can also confirm that it works, thanks @rhklee!

PR #935, opened 2024-02-02
Change PlanGenerator.create_plan to return `UUID` instead of `Plan`

seppeljordan:
Before this commit the data generator class PlanGenerator returned an instance of `Plan`. This was problematic as it coupled the test suite unnecessarily to the data representations internal to the use case logic. This commit addresses the coupling in three ways: 1) We changed the function signature of the PlanGenerator.create_plan method such that we return a UUID instead of an instance of `Plan`. 2) We adjusted all the call sites to match the new signature where this was possible with minimal effort. 3) We created a new method
PlanGenerator.create_plan_record which will be used in instances where we could not simply adjust the call site of PlanGenerator.create_plan.

The newly created PlanGenerator.create_plan_record received a comment to communicate the deprectation of this method.

Plan-ID: b412d2bb-074a-489b-86f6-90b87077c17a (4x)


sloschert (approved):
Thanks for the work! 4xCR

PR #940, opened 2024-02-03
Remove the "toggle availability" completely from the code base

seppeljordan:
Before this change there was a feature built into the app that allowed companies to signal the availability of their products. This feature will be completely removed by this change.


sloschert (approved):
Great. Let me know if I should register consumption.

sloschert:
> @sloschert Thanks for the reminder.
> 
> b412d2bb-074a-489b-86f6-90b87077c17a 1x

1xCR

PR #947, opened 2024-02-10
Run github CI workflow on pull requests

sloschert:
This change should fix the problem that our CI pipeline wouldn't run on pull requests from forks.

Details: https://docs.github.com/de/actions/using-workflows/events-that-trigger-workflows#pull_request

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -1,6 +1,10 @@
 name: Arbeitszeitapp CI Tests
 
-on: push
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
   static-code-analysis:


sloschert:
@seppeljordan I just noticed that this proposed change runs all our checks twice if we push a new branch and create a pull request at the same time. What do you think about if we run the pipeline only on pull request (and using locally the `./run-checks` script when developing)?

sloschert:
> I see two possible solutions to the problems:
> 
> # Ask people to enable github workflows on their forks
> The advantage here would be that we wouldn't bottleneck the computing capacity for our own workflows as the machines allocated for the CI pipeline would count towards that of the fork.
> 
> # Use something like the following:
> ```
> on:
>   push:
>     branches:
>       - master
>   pull_request:
> ```
> 
> The advantage here would be that we would not need to instruct people on how to enable CI on their local fork.

i'd prefer the second solution, but i'm open for other opinions.

PR #948, opened 2024-02-11
Remove nix flake dependency on nixos-23.05

seppeljordan:
This change will remove the dependency on the NixOS 23.05 package set as this version of NixOS is deprecated.  The instances that are operated by us were already migrated to NixOS 23.11 in the past. Discontinuing support for allows us to drop support for python 3.10, therefore allowing us to use python language & library features that were introduced with version 3.11. 

With this we can leverage one of those features by using the newly introduced type `typing.Self`. In the past we had to rely on a third-party library to get same functionality. Now we can drop that reliance on the library typing_extensions and simply use the pythons core library for that purpose. A result of this change is that we require less computing ressources for running our CI setup as there is now one less version of python/nixos that needs to be checked.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -61,10 +61,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - name: Set up Python 3.10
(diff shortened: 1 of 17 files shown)


sloschert (approved):
Great

PR #950, opened 2024-02-11
Fix bug in email confirmation timestamp

seppeljordan:
Before this change we had a bug when it came to confirming member email addresses. The bug was that we would set the email confirmation timestamp for confirmed members unconditionally to datetime.min. This change addresses this bug and sets it to the actual time of confirmation according to the DatetimeService implementation (which relies on datetime.datetime.now).

We also extended the response model for the
GetUserAccountDetailsUseCase to show the time of confirmation for a users email address. This was done to create test for the bug addressed here. In the future this extension could be used to show the email confirmation timestamp to users.

Plan-ID: d21d8863-1179-4b7c-9d17-1701a7f78e76 (1x)

### DIFF
--- a/tests/use_cases/test_get_user_account_details.py
+++ b/tests/use_cases/test_get_user_account_details.py
@@ -1,6 +1,13 @@
+from datetime import datetime, timedelta
 from uuid import uuid4
 
-from arbeitszeit.use_cases import get_user_account_details
+from parameterized import parameterized
+
+from arbeitszeit.use_cases import (
+    confirm_company,
+    confirm_member,
+    get_user_account_details,
+)
 
 from .base_test_case import BaseTestCase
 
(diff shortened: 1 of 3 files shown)


tests/use_cases/test_get_user_account_details.py
```diff
@@ -50,3 +63,83 @@ def test_that_users_email_address_is_shown(self) -> None:
...
+            == expected_timestamp
+        )
+
+    @parameterized.expand(
+        [
+            datetime(2000, 1, 1),
+            datetime(2000, 1, 2),
+        ]
+    )
+    def test_that_confirmation_timestamp_for_company_is_the_time_when_member_account_was_confirmed(
```
sloschert, line 112: typo: "when_COMPANY_account_was_confirmed"

sloschert (approved):
I left one comment (a type in a test function name). Apart from that it looks good. 1xCR

PR #953, opened 2024-02-15
Add password reset request table and repository functions.

rhklee:
- [x] Add database models for `PasswordResetRequest` table.
- [x] Add repositories functions
- [x] Add records
- [x] Migrate database to add new table.


rhklee:
@seppeljordan @sloschert @Amittai-Aviram How are new tables added? Not sure how that works. I added `AUTO_MIGRATE = True` locally, but that didn't seem to do it for my local environment.

sloschert:
> How are new tables added?

Good point, we should document this. You can generate a migration file automatically by running:

`flask db migrate -m "<some message>"`

You can then check and change the automatically generated migration commands in that file. You can test that the migration works as expected by running `flask db upgrade` and `flask db downgrade` (this will migrate your local database). 

If you are satisfied you can commit the migration file.

rhklee:
@seppeljordan @sloschert  My attempt to merge the migration from master failed. Not sure if I should just abandon this branch in fear of making the problem worse.

PR #957, opened 2024-02-15
my_cooperations view: Use Notifier interface

seppeljordan:
EDIT: Add second paragraph. Update planning info

Before this commit the my_cooperations route for companies used a custom system for displaying user notifications in the web frontend. This change modifies the responsible presenter and view template such that the standard notification sytem `arbeitszeit_web.notification.Notifier` is used instead. Because of this change we in the presenter there is no need to handle user notifications explicitly in view template. Instead we can rely on the notification macro used shared by all views that inherit from `base.html`.

The parameters for the ShowMyCooperationsPresenter.present method were made keyword parameters. This change allowed us to make some of those parameters optional. Although this is not very useful for the production code it is however useful to improve the clarity of the test cases. Making `accept_cooperation_response`, `deny_cooperation_response` and `cancel_cooperation_solicitation_response` optional allowed us to not specify those arguments in test cases where we were not interested in the corresponding behavior. This will also be useful in the future, when we overhaul the my_cooperations_view such that there are separate request handlers for the `AcceptCooperation`, `RejectCooperation` and `CancelCooperation` use case.

### DIFF
--- a/tests/www/presenters/test_show_my_cooperations_presenter.py
+++ b/tests/www/presenters/test_show_my_cooperations_presenter.py
@@ -106,15 +106,12 @@ class ShowMyCooperationsPresenterTests(BaseTestCase):
         super().setUp()
         self.presenter = self.injector.get(ShowMyCooperationsPresenter)
 
-    def test_coordinations_are_presented_correctly(self):
+    def test_coordinations_are_presented_correctly(self) -> None:
         presentation = self.presenter.present(
-            LIST_COORDINATIONS_RESPONSE_LEN_1,
-            get_inbound_response_length_1(),
-            AcceptCooperationResponse(rejection_reason=None),
-            DenyCooperationResponse(rejection_reason=None),
-            get_outbound_response_length_1(),
-            None,
-            get_coop_plans_response_length_1(),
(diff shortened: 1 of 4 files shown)


tests/www/presenters/test_show_my_cooperations_presenter.py
```diff
@@ -153,19 +153,14 @@ def test_successfull_accept_request_response_is_presented_correctly(self):
...
-            1,
-        )
-        self.assertEqual(
-            presentation_success.accept_message[0],
-            self.translator.gettext("Cooperation request has been accepted."),
+        assert len(self.notifier.infos) == 1
+        assert self.notifier.infos[0] == self.translator.gettext(
+            "Cooperation request has been accepted."
         )
+        assert not self.notifier.warnings
```
rhklee, line 160: Just a pedantic point, if the warning notification array is checked to be empty/None here shouldn't it be checked in the other tests (but the other way around possibly depending on the case)? Since I suppose it's just as likely in the other cases the other notification array gets inadvertently filled.

rhklee (approved):

rhklee (approved):

sloschert:
> d21d8863-1179-4b7c-9d17-1701a7f78e76

2xCR

PR #958, opened 2024-02-16

seppeljordan:
EDIT: Update description

This commit changes the routine for updating the development dependencies, specifically the arbeitszeit_development.update_constraints script such that more packages are excluded from the constraints.txt file. All packages starting with `types-` will be excluded from now on since there is no benefit in pinning those packages for pip users. The `arbeitszeitapp` itself will be excluded from now on. This was not necessary before this commit since nix handled the arbeitszeitapp as "editable" before. Those packages were already excluded from constraints.txt before this change.


sloschert (approved):
LGTM

Issue #954, opened 2024-02-15
Display additional information for fields in plan details page


petsch0:
When a user creates a new plan (create_draft page), most of the fields to fill out have those small ?-icons in the corner that you can click and then additional information for the field is displayed (not too comfortable to use, but they are there). 

In the overview page for a plan (plan_details page) there is nothing similar for the single fields, however I think this would be very helpful at this place, especially for fields that are calculated (price, labour cost) or generated (plan uuid; what is it good for?) and especially for new users and/or users that do not have strong theoretical background in labour time calculation.

In a way, these context information could serve as point to connect the bare accounting process to the theory on why the accounting is done the way it is done. 

First implementation idea that comes to my mind are those small (i)-icons that will expand to some kind of speech bubble  whenever you hover over them. Later this could even be extended so that when you click on it, a whole new sidebar page opens up above the present one.

petsch0:
I would say yes, and in this context we maybe could also update the additional information that is given in the create_draft page. Maybe we can get some good input for the content from other ida members during next workshop?

PR #959, opened 2024-02-16
Info icon tooltips on plan details page

petsch0:
implements #954 

9d2909df-0681-48ff-a897-495056c78f5d (4x)


petsch0:
@seppeljordan @sloschert since I am not an expert on html and templating, kindly have a look at the initial commit and let me know what you think. This dropdown-thing looks a little hacky to me, but it is comfortable in the UI. Still I think that it would also be okay to re-use the existing info modal from draft_form.html.

sloschert:
I like it, but it seems not to be mobile responsive, because the info text is cut off. Do you think there is a way to make the text fit mobile screens? 

Also, wouldn't be "tooltip" a more accurate name than "dropdown"?

image

sloschert (changes requested):
I left one comment

arbeitszeit_flask/templates/macros/plan_details.html
```diff
@@ -10,6 +11,9 @@ <h1 class="title">
                     <div class="tile">
                         <div class="tile is-parent is-vertical">
                             <article class="tile is-child notification">
+                                <div class="has-text-right has-text-info-dark">
+                                    {{ hover_info_dropdown(id="id_plan_id", text=gettext("This UUID identifies your plan. It is used by workers/companies to debit/credit labour time spent on the plan")) }}
```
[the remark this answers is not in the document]
petsch0: np, thanks for having a look. I changed the text according to your suggestion.

petsch0:
> Do you think there is a way to make the text fit mobile screens?
thanks for checking this. I'll have a look to see if that is possible.

> Also, wouldn't be "tooltip" a more accurate name than "dropdown"?
yes, I changed the PR title, and will also change the file name and the macro.

sloschert:
In thought it would be possible to tab tooltips on mobile, but might be wrong. Sure, we can do it in a future PR (but some effort has been made yet to achieve responsiveness, so we should not forget this). I'll approve and leave it up to @petsch0 how to proceed.

sloschert (approved):

petsch0:
Consider latest texts for various fields as drafts for discussion.

arbeitszeit_flask/templates/macros/plan_details.html
```diff
@@ -58,6 +62,9 @@ <h1 class="title">
                 <div class="tile is-4">
                     <div class="tile is-parent">
                         <article class="tile is-child notification has-background-warning-light">
+                            <div class="has-text-right has-text-info-dark">
+                                {{ info_tooltip(id="id_plan_status", text=gettext("Status can be \"Active\", i.e. plan was approved and current date is within planning timeframe, or \"Inactive\", i.e. plan is not yet approved or it is expired.")) }}
```
petsch0, line 66: can plan be approved but be Inactive and not expired because current date is BEFORE start of planning timeframe?

petsch0:
also as texts are getting longer and longer: do you think existing modals with pop-up windows should rather be re-used instead of those long dropdowns?

sloschert:
> I see that approval is still valid although I pushed some commits. Shouldn't this have been invalidated, maybe with exception of merge commits? 

Apparently we could activate this option in github: https://stackoverflow.com/a/69229270 

but if we activate it we could not allow small cosmetic changes after approval (sometimes we say: "ok change this comment and you can merge"). For now I propose that you simply rerequest a review from us manually.

arbeitszeit_flask/templates/macros/plan_details.html
sloschert: No, currently not.

sloschert:
> also as texts are getting longer and longer: do you think existing modals with pop-up windows should rather be re-used instead of those long dropdowns?

My opinion: I like the popups you implemented more, because they are less intrusive, and I would actually replace the existing modals with these. Regarding the length of the text: I think these tips should be quite short anyway. If they are too long to fit in such a info popup we should rather add this info to a proper user manual (on which we still have to agree, there is a rudimentary one in route `/help`).

arbeitszeit_flask/templates/macros/tooltips.html
```diff
@@ -0,0 +1,23 @@
...
+            <i class="fa-solid fa-info-circle">
+                <span class="icon is-small">
+                    <i class="fas fa-angle-down" aria-hidden="true"></i>
+                </span>
+            </i>
+        </span>
+    </div>
+    <div class="dropdown-menu" id="{{ id }}" role="menu">
+        <div class="dropdown-content">
+            <div class="dropdown-item has-text-left">
```
sloschert, line 14: I find the text a bit too small - what about adding the class "is-size-6"?

arbeitszeit_flask/templates/macros/plan_details.html
```diff
@@ -112,14 +122,23 @@ <h1 class="title">
...
                             <article class="tile is-child notification has-background-danger-light">
+                                <div class="has-text-right has-text-info-dark">
+                                    {{ info_tooltip(id="id_fixed_means", text=gettext("This is sometimes also called depreciation, e.g. if you plan to use a machine that cost 10000 hours to produce, for 10 years, and you have ten one-year-plans, then for this fixed mean of production, the value would be 1000.")) }}
+                                </div>
                                 <p class="title is-4">{{ view_model.details.means_cost[0] }}</p>
                                 <p class="subtitle">{{ view_model.details.means_cost[1] }}</p>
                             </article>
                             <article class="tile is-child notification has-background-danger-light">
+                                <div class="has-text-right has-text-info-dark">
+                                    {{ info_tooltip(id="id_liquid_means", text=gettext("Here, recurring costs that can vary from plan to plan are summarized, e.g. electricity, rent for production halls, maintenance of machines etc.")) }}
```
sloschert, line 133: I think I would propose to delete "that can vay from plan to plan", because also fixed means of production can vary from plan to plan (not every machine will be used for every product). 
 Perhaps we should mention that we refer to marxian "circulating means of production" here? Perhaps we should even rename "liquid" to "circulating". See the wiki page: https://en.wikipedia.org/wiki/Circulating_capital. Not quite sure why we chose "liquid" from the beginning.  

 See Grundprinzipien 1930:
 "As is well known the Marxist category "means of production" comprises machinery and buildings (fixed means of production), and also raw materials and auxiliary materials (circulating means of production). If now we use for fixed means of production the letter p and for the circulating means of production the letter c, then the formula takes on the following form: ( p + c ) + L = product" https://www.marxists.org/subject/left-wing/gik/1930/03.htm
 In German: "Wie bekannt, umfaßt die Marx'sche Kategorie "Produktionsmittel" , Maschinen, Gebäude (feste Produktionsmittel), als auch die Rohmaterialien und Hilfsstoffe (zirkulierende Produktionsmittel). "

 In Grundprinzipien 1935:
 "Each company calculates how much working time is spent on its product, i.e., it determines how many working hours of "fixed" means of production (machines and buildings), how many working hours of "circulating" means of production (raw materials and consumables) and how many directly consumed working hours flow into the product." (p. 111f., https://arbeitszeit.noblogs.org/files/2023/04/GIC-Fundamental-Principles-2.-Ed.1935-1.pdf)

arbeitszeit_flask/templates/macros/plan_details.html
```diff
@@ -133,12 +152,18 @@ <h1 class="title">
             <div class="tile is-ancestor">
                 <div class="tile is-parent">
                     <article class="tile is-child notification">
+                        <div class="has-text-right has-text-info-dark">
+                            {{ info_tooltip(id="id_type", text=gettext("\"Type\" can be either \"Productive\" or \"Public\". Products from the latter can be consumed for free, and are funded via the Factor of Individual Consumption (FIC)-level.")) }}
```
sloschert, line 156: Without "level"?

arbeitszeit_flask/templates/macros/plan_details.html
```diff
@@ -151,6 +176,9 @@ <h1 class="title">
                 </div>
                 <div class="tile is-parent">
                     <div class="tile is-child notification has-background-warning-light">
+                        <div class="has-text-right has-text-info-dark">
+                            {{ info_tooltip(id="id_labour_per_unit", text=gettext("This is to account for real working hours per unit, independent from consumption price which could be reduced, e.g. for public plans.")) }}
```
sloschert, line 180: 1) typo: "is THE account"
2) Actually there are only two reasons why the "price" differs from the costs: Either the plan is part of what we call a "price cooperation" [Grundprinzipien 1930, p. 101: "Die horizontale Zusammenfassung ist zunächst nur eine rechnerische Angelegenheit zur Feststellung der gesellschaftlich durchschnittlichen Produktionszeit und im Zusammenhang damit des Grades der Produktivität der einzelnen Betriebe. Sicher kann und darf es dabei nicht bleiben, sondern es muß auch zu gegenseitiger technischer Durchdringung kommen."] OR the plan is a public plan, then it is zero.
As the info text is now, it appears as if there could be several, and arbitary, reasons why the "price could be reduced".

sloschert:
@petsch0 I left some comments on the info texts for your consideration and perhaps for some future discussions. I think they are a great feature.

arbeitszeit_flask/templates/macros/tooltips.html
petsch0: done

arbeitszeit_flask/templates/macros/plan_details.html
petsch0: done

arbeitszeit_flask/templates/macros/plan_details.html
petsch0: done

sloschert (approved):
Thanks! :-) 4xCR

PR #963, opened 2024-02-22
Remove cooperation data from Plan record

seppeljordan:
Before this change we stored the `cooperation` that a plan was part as a field of the `Plan` record. This was also reflected in the database implementation where `cooperation` was a column in the `plan` table. After this change the `cooperation` attribute is removed from the `Plan` record. Also the DB implementation stores the cooperation data in a new separate table called `plan_cooperation`. The DB interface had to changed slightly because of this, which affected some use cases in a minor way.

Plan-ID: d21d8863-1179-4b7c-9d17-1701a7f78e76 (3x)


sloschert (approved):
Looks good. 3xCR

PR #969, opened 2024-02-28
Move account_a route to users blueprint

sloschert:
With this change, the "A" accounts of companies are accessible to all authenticated users.

Plan: b05bef9c-4dbf-4cd1-9e99-59219dfa9c4b (1x)


sloschert:
@seppeljordan I merge this, could you register the consumption?

Issue #972, opened 2024-03-03
Url in worker invitation mail has relative path


sloschert:
When a company invites a worker, an email gets sent automatically to the worker with a url in order to accept or reject the invitation. Currently this url is relative (e.g. "/member/invite_details/780ba02f-9b3f-4747-b4b3-1aefda50e5ce`) which is not very usefull. 

The call to the `get_answer_company_work_invite_url` method in the `arbeitszeit_flask/url_index.py` module from the `InviteWorkerPresenterImpl` needs to be changed in order to return an absolute url. In flask's `url_for` function this is usually done by setting `_external` to `True` (https://tedboy.github.io/flask/generated/flask.url_for.html). 

Tests should be added (ideally also for similar cases where urls are shown in mails).

sloschert:
Closing this as resolved.

PR #974, opened 2024-03-04
Make url that's shared in a company's member invite external and form…

rhklee:
Fix for #972

plan: fe0d3b00-3fee-47d5-9f46-a1722b4bf873 3x

### DIFF
--- a/arbeitszeit_flask/templates/member/notification-about-work-invitation.html
+++ b/arbeitszeit_flask/templates/member/notification-about-work-invitation.html
@@ -1,7 +1,3 @@
-<p>{{
-    gettext(
-        'You have been invited by a company to join them as a member on Arbeitszeitapp. Accept or reject via this link: "%(invitation_url)s".',
-        invitation_url=invitation_url
-    )
-  }}</p>
+<p>{{ gettext("You have been invited by a company to join them as a member on Arbeitszeitapp. Accept or reject via this link:") }}</p>
+<p><a href="{{ invitation_url }}">{{ invitation_url }}</a></p>
   
--- a/tests/www/presenters/test_show_company_work_invite_details_presenter.py
+++ b/tests/www/presenters/test_show_company_work_invite_details_presenter.py
@@ -28,7 +28,7 @@ class PresenterTests(BaseTestCase):
     def test_successful_response_renders_accept_url_to_view_model(self) -> None:
         invite_id = uuid4()
         expected_url = self.url_index.get_answer_company_work_invite_url(
-            invite_id=invite_id
+            invite_id=invite_id, is_absolute=False
         )
         view_model = self.presenter.render_response(
             self.make_response(invite_id=invite_id)
--- a/arbeitszeit_flask/url_index.py
+++ b/arbeitszeit_flask/url_index.py
@@ -54,9 +54,13 @@ class GeneralUrlIndex:
     def get_company_summary_url(self, company_id: UUID) -> str:
         return url_for("main_user.company_summary", company_id=company_id)
 
-    def get_answer_company_work_invite_url(self, *, invite_id: UUID) -> str:
+    def get_answer_company_work_invite_url(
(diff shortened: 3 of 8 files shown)


arbeitszeit_flask/templates/member/notification-about-work-invitation.html
```diff
@@ -1,7 +1,3 @@
-<p>{{
-    gettext(
-        'You have been invited by a company to join them as a member on Arbeitszeitapp. Accept or reject via this link: "%(invitation_url)s".',
-        invitation_url=invitation_url
-    )
-  }}</p>
+<p>{{ gettext("You have been invited by a company to join them as a member on Arbeitszeitapp. Accept or reject via this link:") }}</p>
+<p><a href="{{ invitation_url }}">{{ invitation_url }}</a></p>
```
Amittai-Aviram, line 2: Good!  But does the link text need to be the bare URL?  Could it be something like "confirm"?  I understand that this is not the goal of this PR.

Amittai-Aviram (commented):
This looks great, but I guess we need to have some sort of test for it before merging with master.  Is that true?

sloschert:
Two remarks and one proposel: 
1) We use this `get_answer_company_work_invite_url` method twice: First, when creating the email sent out to workers (in `InviteWorkerPresenterImpl`), second, when showing the work invite on the worker's dashboard (in `ShowCompanyWorkInviteDetailsPresenter`). Only in the first case this url has to be absolute.
2) How can we test if the url is absolute or relative? As it is now, we could only test this on the flask level with an integration test, because the url_index test implementation on www level (`tests/www/presenters/url_index.py`) does only know about name and arguments of the url_index methods and is converting these into a fake url. 

-> If you change the signature of the method by adding a key word argument "is_absolute", we could 1) distinguish between cases where we need an absolute url and where we don't and 2) could test that the correct argument is passed in the corresponding presenter tests.

Like this:
`def get_answer_company_work_invite_url(self, *, invite_id: UUID, is_absolute: bool) -> str:`

sloschert (changes requested):
Hi, thanks for tackling this issue. I left a suggestion.

arbeitszeit_flask/templates/member/notification-about-work-invitation.html
rhklee: Yeah I think that's a good idea, I'll make an issue for such an enhancement for the emails.

tests/www/presenters/test_show_company_work_invite_details_presenter.py
```diff
@@ -28,7 +28,7 @@ def test_successful_response_renders_to_view_model(self) -> None:
     def test_successful_response_renders_accept_url_to_view_model(self) -> None:
```
rhklee, line 28: Tests the web presenter has the properly parameterized `get_answer_company_work_invite_url` passed.

rhklee:
I added an `is_absolute` path parameter to `get_answer_company_work_invite_url` the presenter tests now check if the `is_absolute` parameter is in the rendered views given how the UrlIndex mock is setup. I guess that seems like the easiest solution.

rhklee:
> LGTM. Let me know if you have a plan ID for registering the consumption of this.

Cool, I just added the plan id above. And here it is again:

plan: fe0d3b00-3fee-47d5-9f46-a1722b4bf873 3x

sloschert:
> > LGTM. Let me know if you have a plan ID for registering the consumption of this.
> 
> @seppeljordan Cool, I just added the plan id above. And here it is again:
> 
> plan: fe0d3b00-3fee-47d5-9f46-a1722b4bf873 3x

Unfortunately this plan has expired yesterday, sorry. Can you create a new one and add to the coop again?

rhklee:
@sloschert new plan-id: `adc48617-8e5e-4cf5-8eb6-09be373428cf`

sloschert:
Consumption registered (3x)

Issue #964, opened 2024-02-25
Replace info modals with info dropdowns


petsch0:
follow-up from #954. see comment here https://github.com/arbeitszeit/arbeitszeitapp/pull/959#issuecomment-1956183467

Issue #412, opened 2022-06-04
Improve input labels and help text on plan creation page


sloschert:
**Description**

The create plan draft page (`/company/create_draft`) should be more user friendly by explaining what fixed and liquid means of production are  and which units are requested in the user input fields

**Details:**

As remarked by several users in our workshop the following things should be improved on the  view:

- [x] Explain what fixed and liquid means of production are. This information could be given in the "help modal" (the blue circle with a question mark in it) with the id `id_costs`.
![Bildschirmfoto vom 2024-02-05 21-38-19](https://github.com/arbeitszeit/arbeitszeitapp/assets/61537351/d353b894-9f29-480d-b092-aa6923160ae6)

- [x] Make it explicit that users should specify the following input fields in hours: "Costs for fixed means of production", "Costs for liquid means of production", "Costs for labour". This could be done for example by adding "hours" in parenthesis ("Costs for fixed means of production (hours)", "Costs for liquid means of production (hours)", "Costs for labour (hours)".)

petsch0:
second task was done in #973

petsch0:
first task done in #975

PR #975, opened 2024-03-09
Tooltips on plan creation page

petsch0:
#412 #964

9d2909df-0681-48ff-a897-495056c78f5d (2x)


arbeitszeit_flask/templates/macros/draft_form.html
```diff
@@ -39,18 +36,21 @@
 </div>
 <div class="block py-2"></div>
 <div class="box has-background-info-light">
-  <div class="has-text-right has-text-info-dark js-modal-trigger" data-target="id_unit_amount">
-    <span class="icon"><i class="fa-solid fa-circle-question"></i></span>
-  </div>
   <div class="field">
+    <div class="has-text-right has-text-info-dark">
+      {{ info_tooltip(id="id_unit", text=gettext("This value is used for calculation purposes only, i.e. you do not have to adhere to any rules here. For example, if you plan to produce beverages, you are free to choose if your smallest delivery unit is one can or one six-pack of cans."))}}
```
sloschert, line 41: It is actually more of an "informational purpose", I would say (you inform the potential consumers). It is, by the way, also relevant for cooperations, where the delivery units should match (we do not enforce this, though). 

What about this?

"This value is used for informational purposes only, i.e. you do not have to adhere to any rules here. For example, if you plan to produce beverages, your smallest delivery unit might be one can or one six-pack of cans."

arbeitszeit_flask/templates/macros/draft_form.html
```diff
@@ -39,18 +36,21 @@
...
     <div class="control">
       {{ form.prd_unit(class_="input is-large") }}
-      <p class="help">{{ gettext("E.g. 1 package of 100 pieces, 1 kilogram, 1 language lesson of 1.5 hour, etc.")}}</p>
+      <p class="help">{{ gettext("E.g. 1 package, 1 kilogram, 20 bottles, 1 language lesson, etc.")}}</p>
     </div>
   </div>
   <div class="block"></div>
   <div class="field">
+    <div class="has-text-right has-text-info-dark">
+      {{ info_tooltip(id="id_amount", text=gettext("This value is assumed to be in the same unit as Smallest delivery unit. By dividing the sum of your production costs through this amount, the app calculates how much labour time is required to produce one piece of your product."))}}
```
sloschert, line 52: "Smallest" -> "smallest"
"divide BY"

What about changing the first sentence?

'How many units, as described in "smallest delivery unit", will you produce? By dividing the sum of your production costs by this amount, the app calculates how much labour time is required to produce one piece of your product.'

sloschert (changes requested):
Great, thanks for the change. I propose some changes, please have a look if this makes sense for you. 

(Not in the scope of this Pull Request): In the last pull requests we added quite some additional (English) text. We should not wait too long with translating these texts into German. If you are interested in an intro into the translation pipeline let me know, we can have a call or meet for that purpose.

arbeitszeit_flask/templates/macros/draft_form.html
petsch0: done. The info about cooperations is maybe sth for the "help" docu?

arbeitszeit_flask/templates/macros/draft_form.html
petsch0: done, changed first sentence to "How many of your \"smallest delivery unit\" will you produce in total?"

arbeitszeit_flask/templates/macros/draft_form.html
sloschert: Yes, the whole cooperation thing is still missing there.

sloschert:
Thanks. If you want to, let us know your plan and how long it took.

sloschert (approved):

sloschert:
2x consumption registered

PR #979, opened 2024-03-13
Non-colored balances of company accounts

sloschert:
Before this commit balances of company accounts p, r, a and prd were shown green when value was >= 0 and red when <0. This was misleading, because it potentially shows a "green light" for misplanning (less consumption of means of production, less worked hours, more product delivered than originally planned, etc.). The real goal of planning is to be near zero. 

The company balances are now always non-colored. Green and red coloring has been maintained for transactions (it's easier in order to distinguish between positive and negative numbers) and for the member account (here it is more probable that it is problematic if a worker consumes more than the amount of available work certificates).

Plan: b05bef9c-4dbf-4cd1-9e99-59219dfa9c4b

### DIFF
--- a/arbeitszeit_flask/templates/user/company_account_a.html
+++ b/arbeitszeit_flask/templates/user/company_account_a.html
@@ -22,7 +22,7 @@
         </div>
         <p>{{ gettext("Balance:") }}</p>
         <p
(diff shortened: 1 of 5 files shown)


rhklee (approved):

sloschert:
1xCR

PR #981, opened 2024-03-29
Fix bug in QueryPlansView

sloschert:
In the QueryPlansView we were passing `request.form` instead of `request.args` to the PlanSearchForm. This was incorrect because we are dealing with a GET request and key word arguments.

The bug has been fixed and integration tests have been added.

Plan: b05bef9c-4dbf-4cd1-9e99-59219dfa9c4b (2x)


rhklee (approved):

sloschert:
2xCR

PR #991, opened 2024-04-10
Don't generate transactions for 0 fixed or liquid means planned

### LINKED ISSUE #337: Transactions of value zero are shown in account views of companies
Transactions of value zero are shown currently in account views of companies. This transactions of value 0 occur when credit for p, r or a is paid out for a plan where p, r, o a are zero.  

While this does not anything severe, it does clutter unnecessarily the account views.

Does the rest agree that this is unintended behaviour?

I see two options of solving this:
1) Do not create a transaction of zero in the first place
2) Hide transactions of zero in the account views 

I prefer the second solution

Example ("p account"):
image

seppeljordan:
Before this commit the application would generate transactions with a volume of 0 hours when plans did not include any fixed or liquid means planned. This commit only generates those transactions in cases where the planning company actually plans to use fixed or liquid means.

fixes #337

### DIFF
--- a/arbeitszeit/use_cases/approve_plan.py
+++ b/arbeitszeit/use_cases/approve_plan.py
@@ -35,12 +35,14 @@ class ApprovePlanUseCase:
         matching_plans.update().set_approval_date(now).set_activation_timestamp(
             now
         ).perform()
-        self._create_transaction_from_social_accounting(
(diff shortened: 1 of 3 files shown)


sloschert (approved):
thanks, let me know the plan id if you want to.

PR #994, opened 2024-04-11
Introduce JsonList to possible json schema definition types

seppeljordan:
Before this commit the schema definition for json endpoints had embedded the concept of a list in json into all the other schema types by having the attribute `as_list`. This attribute was completely removed. Instead we introduced a dedicated type `JsonList` that describes the fact that a certain value should be a list of values. Unfortunately this change broke all the schema tests in our API code that relied on the `as_list` feature. All schema tests that broke were removed without a replacement. We did not bother with a replacement because it is clear the API subsystem needs a complete re-design as it relies on deprecated libraries, an awkward interface an a schema definition system that is completely decoupled from the values that are actually produced by the API endpoints.

Plan-ID: d21d8863-1179-4b7c-9d17-1701a7f78e76 (3x)

### DIFF
--- a/arbeitszeit_flask/api/schema_converter.py
+++ b/arbeitszeit_flask/api/schema_converter.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Union
+from typing import TypeAlias, Union
 
 from deepdiff import DeepDiff
 from flask_restx import Model, fields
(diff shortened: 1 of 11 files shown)


sloschert (approved):
3xCR

PR #997, opened 2024-04-12
Remove deepdiff python package from dependencies

seppeljordan:
Before this commit we would use the deepdiff library to check if an API schema with a given name was already registered.  We would throw an error if the schema registered was any different from the one we were about to register.  This was done to prevent accidental overriding of registered schemas.  We switched our appoach and now check for schema identify, meaning that we only registering schemas under the same name if they are infect the same object. This made the usage of deepdiff obsolete. Consequently the library was removed from our dependency list.


sloschert (approved):
2xCR

PR #998, opened 2024-04-12
Regenerate type stubs

seppeljordan:
After the latest update of `nixpkgs` we need to create the type stubs for our python dependencies again.  This is achieved by this commit. Further the type_stubs are no subject to automatic code formatting to avoid unnecessary large diffs should mypy change the code formatting of the generated type stubs for some reason.


sloschert (changes requested):
You write: "Further the type_stubs are no subject to automatic code formatting", but `type_stubs/` is in the file `.autoformattingrc`?

sloschert (approved):

PR #1001, opened 2024-04-16
Add implicit "/company" path prefix to company routes

seppeljordan:
In turn we removed the explicit /company prefix of all routes in the company blueprint.


sloschert (approved):
1xCR

PR #1004, opened 2024-04-16
Simplify html template for draft creation

seppeljordan:
Before this commit the web logic around draft creation was slightly
complicated. The complication was that the "Cancel" action in the
creation form was handled as a POST request that would simply redirect
the user to another page. Now the same thing is achieved via a `<a>`
element. This allowed us to simplify the route/view.

### DIFF
--- a/arbeitszeit_flask/templates/company/create_draft.html
+++ b/arbeitszeit_flask/templates/company/create_draft.html
@@ -20,22 +20,15 @@
         <div class="field is-grouped is-grouped-centered">
           <div class="control">
             <button class="button is-primary is-light"
-                    name="action"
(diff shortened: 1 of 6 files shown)


sloschert (changes requested):
Manually testing I think I found a problem in this commit: Before this change you could "edit" a draft from the "my plans" page (by clicking on the edit icon) and cancel the editing by clicking the "cancel" button. No new draft got created. 
Having this branch checked out, a click on "cancel" does create a copy of the draft. So after clicking "cancel" you have now one more drafts in your list of drafts.

sloschert (approved):

PR #1005, opened 2024-04-16
Make company accounts info available to all authenticated users

seppeljordan:
This commit makes the account data for all companies available to all authenticated users. This was achieved by linking to the already implemented url routes in the company summary view.


sloschert:
I think the navigation is currently a bit confusing, when you are viewing the account of a different company, because currently there is no information on the fact that it is not your own account. Also when you use the "tab" like navigation element that says "Accounts" you get navigated to the accounts overview of the foreign company, but again there is no hint that it's not your accounts. 

I'd be ok with merging it anyway now, but we should remember to improve this.

image

sloschert (approved):
1xCR

PR #1006, opened 2024-04-17
Include DB tests in nix build process

seppeljordan:
With this commit all tests from the test suite are executed as part of the nix build process. Previously only the tests that did not require a running instance of postgresql were executed. Now the DB tests are also run.

# Implications

If we would merge this then we would have two checks less in our github pipeline (The CI feature of GitHub). On the other hand the `flake-check` pipeline would take more time to finish.

With the current implementation we would be able to run the tests (including the DB tests) on any machine that has `nix` available regardless of the local postgres setup. This would make all tests reproducible with regards to the DB.

# Possibilities

A similar approach would allow us to have a DB setup that "just works" (TM) on any machine that runs `nix`.

# MacOS
Please note that this change was only tested on `x86_64-linux`. 

@Amittai-Aviram @petsch0 @rhklee It would be great if you could test this on MacOS if you have the means and the time for it. All you have to do is check out this branch and run the nix checks like so (assuming that `origin` points at this repository):

```
git fetch
git checkout origin/seppeljordan/db-tests-in-nix
nix flake check --print-build-logs
```

Afterwards you can post the build output here if the `nix flake check` command fails or write me an email with the content of the logs. The build might take some time since you might need to download some packages for the stable version of `nixpkgs`.


petsch0:
Hi @seppeljordan, not sure what exactly you need from the logs, therefore attaching the whole output. As far as I can tell the command succeeded as it was supposed to.
[nix-flake-check-build-logs-darwin.txt](https://github.com/ida-arbeitszeit/arbeitszeitapp/files/15120030/nix-flake-check-build-logs-darwin.txt)

sloschert (approved):
2xCR

PR #1009, opened 2024-05-04
Rename GetPrefilledDraftDataPresenter to GetDraftDetailsPresenter

seppeljordan:
This GetPrefilledDraftDataPresenter was confusingly named as it was really the presenter responsible for rendering draft details.  This issue was addressed by renaming the presenter in question and the associated modules and tests.


sloschert (approved):
1xCR

PR #1014, opened 2024-05-05
Implement plugin system for EmailService.

seppeljordan:
This commit introduces a plugin system for the EmailService in the arbeitszeitapp. The new system empowers administrators to customize email delivery methods according to their requirements. Notably, this enhancement allows for independent testing of email functionality, facilitating the eventual removal of the dependency on `flask_mail` from our codebase.


sloschert:
Thanks for working on this. A few questions: 

- Out of curiosity: What has still to be done before we can remove flask_mail?
- There is a section in our Readme, called "E-mail Configuration". Can this get deleted, because we have the development guide, or is it still necessary in your opinion?
- I noticed that in the "development_guide.rst", there is no description on how to implement an "email plugin". It only says "This option must be a python module path to the email plugin to be used." On purpose?

sloschert (approved):
2xCR

PR #1016, opened 2024-05-05
Update integration tests to use MockEmailService instead of flask_mail.

seppeljordan:
Previously, we used `flask_mail` for testing email notifications. Now,
we've created `MockEmailService`, designed with a similar interface to
the testing interface of `flask_mail`, and integrated it with the new
`MailPlugin`. This change makes our flask integration tests no longer
dependent on `flask_mail`, paving the way to remove flask_mail from
our codebase in the future.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,9 +19,12 @@ include-package-data = true
 
 [tool.setuptools.packages.find]
 include = [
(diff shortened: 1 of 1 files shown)


sloschert (approved):
1xCR

PR #1017, opened 2024-05-05
Remove dependency on flask_mail

seppeljordan:
With this commit we removed the dependency on `flask_mail`. Instances
that are configured to use `flask_mail` as their mail backend will
automatically use `SmtpMailService` instead.


sloschert (approved):
1xCR

PR #1020, opened 2024-05-11
Delete old email address after email change

sloschert:
Before this change it was not possible for a user, after changing its email address, to change back to its original email address. More generally, an email address no longer in use could not be reused.

This was fixed by deleting the old email address from our records.

Plan: b05bef9c-4dbf-4cd1-9e99-59219dfa9c4b (2x)


sloschert:
New Plan: 371825c0-3112-47cd-9cb0-fea799fad3ee

PR #1021, opened 2024-05-11
Translate app to Spanish

sloschert:
This PR updates German translations, adds support for Spanish and translates all strings to Spanish. As documented in the README, I found a gettext extension for VS Code that was very helpful in combination with a LLM. While a Spanish native speaker might find some smaller issues in the translations, it should be good enough for now.  

Commits:

- Update pot and po files
- Update translations for German language
- Add Spanish language support
- Translate app to Spanish

Plan: b05bef9c-4dbf-4cd1-9e99-59219dfa9c4b (4x)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -404,7 +404,8 @@ for translation.
 
 #. Translate language-specific ``.po`` files. For translation
    programs, see `this page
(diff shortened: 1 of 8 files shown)


sloschert:
New Plan: 371825c0-3112-47cd-9cb0-fea799fad3ee

PR #1025, opened 2024-05-25
Implement proper draft editing

seppeljordan:
This commit implements proper draft editing for companies.  Before this commit companies would create a new draft everytime they tried to edit an existing draft. This was due to the EditDraft use case not being called. Instead we would call to the CreatePlanDraft use case every time.

Plan-ID: d21d8863-1179-4b7c-9d17-1701a7f78e76 (4x)

### DIFF
--- a/arbeitszeit_web/www/controllers/edit_draft_controller.py
+++ b/arbeitszeit_web/www/controllers/edit_draft_controller.py
@@ -1,0 +1,89 @@
+from dataclasses import dataclass
+from uuid import UUID
+
+from arbeitszeit.use_cases.edit_draft import Request
+from arbeitszeit_web.forms import DraftForm
+from arbeitszeit_web.notification import Notifier
+from arbeitszeit_web.session import Session
+from arbeitszeit_web.translator import Translator
+
+
+@dataclass
+class EditDraftController:
+    notifier: Notifier
+    translator: Translator
+    session: Session
+
+    def process_form(self, form: DraftForm, draft_id: UUID) -> Request | None:
+        product_name = form.product_name_field().get_value()
+        if not product_name:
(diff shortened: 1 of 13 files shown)


arbeitszeit_web/www/controllers/edit_draft_controller.py
```diff
@@ -0,0 +1,89 @@
...
+class EditDraftController:
+    notifier: Notifier
+    translator: Translator
+    session: Session
+
+    def process_form(self, form: DraftForm, draft_id: UUID) -> Request | None:
+        product_name = form.product_name_field().get_value()
+        if not product_name:
+            form.product_name_field().attach_error(
+                self.translator.gettext("The product name field cannot be empty.")
```
sloschert, line 21: Just FYI: These error messages currently are not visible to the user (at least in the browsers I tested). Instead, a message like "Fülle dieses Feld aus" (Chrome) or "Bitte füllen Sie dieses Feld aus" (Firefox) appears.

tests/www/controllers/test_edit_draft_controller.py
```diff
@@ -0,0 +1,318 @@
...
+        self.controller = self.injector.get(EditDraftController)
+        self.session.login_company(uuid4())
+
+    @parameterized.expand(
+        [
+            "test name",
+            "other test name",
+        ]
+    )
+    def test_that_product_name_from_description_is_used_in_request(
```
sloschert, line 23: i think it should be "...product_name_from_form..."

sloschert (changes requested):
Great change, thanks. If found an unexpected behaviour.

sloschert (approved):

sloschert:
4xCR

PR #1028, opened 2024-06-02
Refactor test_list_outbound_coop_requests.py into class based suite

seppeljordan:
Before this change the test suite in
tests/use_cases/test_list_outbound_coop_request.py was still using the pytest function style.  This commit refactors those tests to fit the usual class based style.


sloschert (approved):
1xCR

PR #1030, opened 2024-06-17
Show workplace emails as mailto link

sloschert:


Amittai-Aviram (approved):
Nice!  Is this change small enough not to require a test?

Issue #728, opened 2023-05-27
New Feature: Change member email address


seppeljordan:
Members should be able to change their email addresses. The following steps describe how this process should be handled:

1. Users should be able to request that their email address should be changed via the web interface. When requesting this change the user must provide the new email address in a web form. The system will reject the request if the email is already in use by any other account. 
2. If accepted then the user will receive an email with a confirmation link to their new email address. At the same time the system will send an email to the old email address, warning the user that the process to change their email address in the arbeitszeitapp was initiated. This email should contain an email address that lets the user contact an admin in case someone else is messing with their account. 
3. The confirmation link should lead to a page where the user can confirm the change. The user must provide their password for authentication purposes in this form. 
4. After the change was confirmed the system will use the new email address and "forget" about the old email address.

These steps were discussed in #702

sloschert:
@seppeljordan Just to be sure: I will implement this in step 3, ok? "The user must provide their password for authentication purposes in this form."

PR #1031, opened 2024-06-18
Create ChangeUserEmailAddressPresenter

sloschert:
This presenter is intended to be used in the `change_email_address` POST route in `arbeitszeit_flask/user/routes.py` and presents the response of the `ChangeUserEmailAddressUseCase`.

Contributes to #728.

Plan-ID: 371825c0-3112-47cd-9cb0-fea799fad3ee (2x)

### DIFF
--- a/arbeitszeit_web/www/presenters/change_user_email_address_presenter.py
+++ b/arbeitszeit_web/www/presenters/change_user_email_address_presenter.py
@@ -1,0 +1,34 @@
+from dataclasses import dataclass
+
+from arbeitszeit.use_cases import change_user_email_address
+from arbeitszeit_web.notification import Notifier
+from arbeitszeit_web.translator import Translator
+from arbeitszeit_web.url_index import UrlIndex
+
+
+@dataclass
+class ViewModel:
+    redirect_url: str | None
+
+
+@dataclass
+class ChangeUserEmailAddressPresenter:
+    notifier: Notifier
+    url_index: UrlIndex
+    translator: Translator
+
+    def render_response(
+        self, response: change_user_email_address.Response
+    ) -> ViewModel:
+        if response.is_rejected:
+            self.notifier.display_warning(
+                self.translator.gettext(
+                    "Something went wrong. Perhaps the email address is not valid or already in use."
+                )
+            )
+            return ViewModel(redirect_url=None)
+        else:
+            self.notifier.display_info(
+                self.translator.gettext("Email address changed successfully.")
+            )
+            return ViewModel(redirect_url=self.url_index.get_user_account_details_url())
--- a/tests/www/presenters/test_change_user_email_address_presenter.py
+++ b/tests/www/presenters/test_change_user_email_address_presenter.py
@@ -1,0 +1,52 @@
+from arbeitszeit.use_cases import change_user_email_address
+from arbeitszeit_web.www.presenters.change_user_email_address_presenter import (
+    ChangeUserEmailAddressPresenter,
+)
+from tests.www.base_test_case import BaseTestCase
+
+
+class ChangeUserEmailAddressPresenterTests(BaseTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.presenter = self.injector.get(ChangeUserEmailAddressPresenter)
+
+    def test_that_view_model_contains_redirect_url_to_account_details_when_response_is_not_rejected(
+        self,
+    ) -> None:
+        response = self.create_response(is_rejected=False)
+        view_model = self.presenter.render_response(response)
+        assert view_model.redirect_url == self.url_index.get_user_account_details_url()
+
+    def test_that_view_model_contains_no_redirect_url_when_response_is_rejected(
+        self,
+    ) -> None:
+        response = self.create_response(is_rejected=True)
+        view_model = self.presenter.render_response(response)
+        assert view_model.redirect_url is None
+
+    def test_that_a_warning_is_displayed_when_response_is_rejected(self) -> None:
+        response = self.create_response(is_rejected=True)
+        assert not self.notifier.warnings
+        self.presenter.render_response(response)
+        assert self.notifier.warnings
+
+    def test_that_no_warning_is_displayed_when_response_is_not_rejected(self) -> None:
+        response = self.create_response(is_rejected=False)
+        assert not self.notifier.warnings
+        self.presenter.render_response(response)
+        assert not self.notifier.warnings
+
+    def test_that_no_info_is_displayed_when_response_is_rejected(self) -> None:
+        response = self.create_response(is_rejected=True)
+        assert not self.notifier.infos
+        self.presenter.render_response(response)
+        assert not self.notifier.infos
+
+    def test_that_an_info_is_displayed_when_response_is_not_rejected(self) -> None:
+        response = self.create_response(is_rejected=False)
+        assert not self.notifier.infos
+        self.presenter.render_response(response)
+        assert self.notifier.infos
+
+    def create_response(self, is_rejected: bool) -> change_user_email_address.Response:
+        return change_user_email_address.Response(is_rejected=is_rejected)


4lm:
👀 I'll have a look at this PR.

4lm (approved):
I like what I see and approve of this PR. It is a pleasure to look at well-structured code like this (not only in this PR but in the whole code base).

What I looked at:

- The presenter that creates the ViewModel implements the two response cases: reject and success. These two cases are all necessary response cases for this use case. Therefore, the presenter seems well-formed.
- The seven tests of  the presenter cover all test cases I also could think of:
   - reject -> warning, yes
   - reject -> info, no
   - reject -> ViewModel, url_index,  no
   - success -> warning, no
   - success -> info, yes
   - success -> ViewModel, url_index, yes (user account details URL)

I have some minor remarks that are no merge-stoppers and can be discussed or treated here but also elsewhere so as not to jam the merge flow:

-  `url_index` is either of type `None` or `str`. In the rejection case, the test is only truthy but not explicit: `assert not x` (truthy) vs. `assert x is None` (explicit). To be a little fuzzy here can be by intention because the ViewModel should only consist of string and bools (if I remember Uncle Bob's intention with the ViewModel approach correctly). Strictly viewed, the None type should then be an empty string (rejection) vs a well-formed URL string (success) and shouldn't use type None at all - right?
-  The creation of the URL string is lifted to the outer flask layer, which is the right thing to do, following the Clean Architecture approach. But there seems to be no check in any layer if the URL strings are well formed. Only the trust that the outer (framework) layer handles URL creation correctly. Looking at the Flask layer, `url_for` returns `Any` via the linter. Therefore, the URL string implementation seems a little shaky. Or am I missing something here?
- The live app has EN, ES, and DE, but there are no corresponding .po file entries for this PR's warning and info strings. Or is this normally done in a different PR, e.g., by a translator (different person)?
- The first of the two tests that test `url_index` could IMO be removed because the second test, which tests that `url_index` is correct, inherently has the requirement for `url_index` to exist in order to pass.

What about merging? Is this done by the request person or the reviewer in this project?

sloschert:
Hey, thanks for your review and remarks! 
1) Before I'll have a look at your remarks: Merging is done by the requesting person after approval. 
2) I think it's just a mistake that you don't have write access. We can talk about this later today.

arbeitszeit_web/www/presenters/change_user_email_address_presenter.py
```diff
@@ -0,0 +1,34 @@
...
+
+from arbeitszeit.use_cases import change_user_email_address
+from arbeitszeit_web.notification import Notifier
+from arbeitszeit_web.translator import Translator
+from arbeitszeit_web.url_index import UrlIndex
+
+
+@dataclass
+class ViewModel:
+    redirect_url: str | None
```
Amittai-Aviram, line 11: Lexi brings up a good point here.  What is the rationale for using the union type str | None rather than simply str and having the empty string mean "no URL"?

arbeitszeit_web/www/presenters/change_user_email_address_presenter.py
```diff
@@ -0,0 +1,34 @@
...
+    url_index: UrlIndex
+    translator: Translator
+
+    def render_response(
+        self, response: change_user_email_address.Response
+    ) -> ViewModel:
+        if response.is_rejected:
+            self.notifier.display_warning(
+                self.translator.gettext(
+                    "Something went wrong. Perhaps the email address is not valid or already in use."
```
Amittai-Aviram, line 26: I guess it would be ideal if the diagnosis were more specific.  Would the user be able to contact a system administrator to find out exactly what went wrong?  Also: will the request be rejected if the user's requested new e-mail address is the same as the old one?

Amittai-Aviram (approved):
Great!  Please see questions.

arbeitszeit_web/www/presenters/change_user_email_address_presenter.py
sloschert: As discussed yesterday I will merge this as-is. We can keep this in mind and discuss later if desired.

arbeitszeit_web/www/presenters/change_user_email_address_presenter.py
sloschert: This would need some differentiation on the use case side (`arbeitszeit/use_cases/change_user_email_address.py`). I did not see this as a priority to be honest. If you do, just comment here and I will fix it in an upcoming PR.

So far we have not discussed, if, when and how users should be able to contact a system admin. A good point for one of the next meetings IMO. 

Yes, when the requested new email exists, the request will be rejected.

sloschert:
@4lm 

- I changed the test to `assert view_model.redirect_url is None` and deleted the first test of `url_index` (forced update)
- Regarding testing the url creation in flask: Have you seen the module `tests/flask_integration/test_url_index.py`? Does this answer your question? Moreover: `url_for` returns `str` in my nix setup.
- translations: Yes, translations are updated every now and then in different PRs. Until then the strings are available only in English.

sloschert:
@seppeljordan I registered 2x consumption

4lm:
> @4lm: ... the ViewModel should only consist of string and bools (if I remember Uncle Bob's intention with the ViewModel approach correctly). Strictly viewed, the None type should then be an empty string ... ?

I'm totally OK with the merge but want to clarify my point further because I believe I didn't get the point across here, and yesterday in our meeting (I was really tired, and my verbal communication skills seemed to be on vacation without giving me a proper heads up).

My "minor" issue with using "None" is only specific to the ViewModel in a Clean Architecture™ context. The ViewModel properties returned by the presenter should only consist of strings, bools, and structural data types (think List, Set, Dict, or @dataclass in a Python context) holding those strings and bools. They also should not be union types, as @Amittai-Aviram correctly noted. The ViewModel's intended use is in the outer framework (Flask) layer for rendering the UI, in our case, HTML via Flask, even if Flask sanitizes "None" correctly to an empty string ("). The data contained in the VideoModel (as I understood Uncle Bob and @seppeljordan in the past) should be as agnostic and basic as possible and shouldn't leak programming language-specific types as data.

I looked through the source code, and I think the implementation in [`list_all_cooperations_presenter.py`](https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/master/arbeitszeit_web/www/presenters/list_all_cooperations_presenter.py) is a good representation of what I mean.

sloschert:
> > @4lm: ... the ViewModel should only consist of string and bools (if I remember Uncle Bob's intention with the ViewModel approach correctly). Strictly viewed, the None type should then be an empty string ... ?
> 
> 
> 
> I'm totally OK with the merge but want to clarify my point further because I believe I didn't get the point across here, and yesterday in our meeting (I was really tired, and my verbal communication skills seemed to be on vacation without giving me a proper heads up).
> 
> 
> 
> My "minor" issue with using "None" is only specific to the ViewModel in a Clean Architecture™ context. The ViewModel properties returned by the presenter should only consist of strings, bools, and structural data types (think List, Set, Dict, or @dataclass in a Python context) holding those strings and bools. They also should not be union types, as @Amittai-Aviram correctly noted. The ViewModel's intended use is in the outer framework (Flask) layer for rendering the UI, in our case, HTML via Flask, even if Flask sanitizes "None" correctly to an empty string ("). The data contained in the VideoModel (as I understood Uncle Bob and @seppeljordan in the past) should be as agnostic and basic as possible and shouldn't leak programming language-specific types as data.
> 
> 
> 
> I looked through the source code, and I think the implementation in [`list_all_cooperations_presenter.py`](https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/master/arbeitszeit_web/www/presenters/list_all_cooperations_presenter.py) is a good representation of what I mean.
> 
> 

Thanks for raising this again, I agree. Soon I will push the Pull Request with the broader context (the usage of the presenter in the flask layer/view function), then we will have an opportunity to discuss this further or rectify it.

4lm:
> * Regarding testing the url creation in flask: Have you seen the module `tests/flask_integration/test_url_index.py`? Does this answer your question? Moreover: `url_for` returns `str` in my nix setup.

Yes, this does answer it — thank you for clarifying! Also, please note that my remark concerning well-formed URL string was a little brainfart because the framework (Flask) layer is not the interface and adapter layer in our project, aka the "web" layer of the presenter. Therefore, something like "url_index" must be a string (as the data type and not a specific URL encoded string type to be agnostic), and the outer Flask layer is responsible for creating well-formed URLs from those strings; what `url_for` does, and what is tested by the tests you are mentioning.

Something seems incorrect with my `nix develop` setup in combination with VS Code. `url_for` returns Any because Flask is not imported in VS Code. Therefore, IntelliSense is not properly working. So my remark on  "`url_for` returns Any" was a false positive.

PR #1032, opened 2024-06-18
Create ChangeUserEmailAddressController

sloschert:
This controller is intended to be used in the `change_email_address` route in `arbeitszeit_flask/user/routes.py`. It has two methods. One, `extract_email_addresses_from_token` is used in the GET and the POST method of the route, the second method, `create_use_case_request`, is used in the POST method and creates the UseCaseRequest object of the `ChangeUserEmailAddressUseCase`.

Contributes to #728.

Plan-ID: 371825c0-3112-47cd-9cb0-fea799fad3ee (2x)


4lm (commented):
I like what I see here in principle, but I have a few comments I want to discuss.

The systems I know concerning email address change normally have a timeout (allowed time delta) of 5 to 15 minutes, not one day. I think the timeout of 1 day should, therefore, be changed to 15 minutes max (context, not of concern in this PR but maybe important for later: for better UX, these systems often directly add a "type in your password" field next to the "type new email" field, even if the user is logged in).
- `extract_email_addresses_from_token` validates the token via `confirm_tolen` in the token service and returns the split ":" but doesn't validate the shape of the provided email (if it is well-formed and valid*). If you have planned to do this in a different PR, good; if not, you maybe should do so.

The unit of work in this PR may be too small. For example, the outer Flask layer for `ChangeUserEmailAddressController` is not implemented, so it's not easy for a reviewer new to the project -me :) - to know how the complete implementation of that controller will look, resulting in the questions above.

*People sometimes use throw-away emails, such as "john.doe+moreletters@gmail.com" for an email originating from "john.doe@gmail.com." Providers like Gmail ship this feature to generate throw-away emails; some call it email alias. Are these alias emails valid replacement emails, or should they be handled as the origin email? Or can this case be ignored but not forgotten for now?

Amittai-Aviram (approved):

sloschert:
Hey, 
regarding your comments:

- I will reduce the timeout to 15 minutes then.
- We have the function `is_possibly_an_email_address` in `arbeitszeit/email.py` and we make use of it for checking the well-formedness of the email-address in the use case `ChangeUserEmailAddressUseCase`. So the controller in this PR creates the request object for the use case where the email is then checked for well-formedness.
-  We have no logic to handle the mentioned emails aliases for now. I have no strong opinion about it. let me know if you think that we should handle it now.
- We agreed more or less to keep the unit of work as small as possible, but I can make it a bit larger to make the bigger picture more visible. Or push PR drafts with upcoming PRs, that should help as well.

4lm (approved):
TL;DR: I approve this PR as is.

Context:

> - I will reduce the timeout to 15 minutes then.

OK.

> - We have the function is_possibly_an_email_address in arbeitszeit/email.py and we make use of it for checking the well-formedness of the email-address in the use case ChangeUserEmailAddressUseCase. So the controller in this PR creates the request object for the use case where the email is then checked for well-formedness.

I understand. Thanks for clarifying.

> - We have no logic to handle the mentioned emails aliases for now. I have no strong opinion about it. let me know if you think that we should handle it now.

Neither do I. Let's move on from this question. We can act on it later if we notice a need to.

> - We agreed more or less to keep the unit of work as small as possible, but I can make it a bit larger to make the bigger picture more visible. Or push PR drafts with upcoming PRs, that should help as well.

Good suggestions! Do as you please; the unit of work thing might only be a problem for new reviewers like me. Or maybe an issue with checkboxes (issues as a checklist) for all upcoming PRs concerning "one" feature (use case) implementation could be the solution. Many open draft PRs can sometimes be confusing if multiple people work on the same project.

sloschert:
@seppeljordan I register 2x consumption

PR #1033, opened 2024-06-30
Change user email address (Flask layer)

sloschert:
**Add change email functionality to user account details page** 

This commit adds the functionality to change the email address on the user account details page. It includes changes to the `account_details.html` template, `url_index.py`, `user_account_details_presenter.py`, and related tests. The user can now click on the "Change email address" link to navigate to the email change page.

**Implement email address change in flask layer** 

This commit implements the email address change flask view. It nearly completes feature request #728.

Only step 3 ("The user must provide their password for authentication purposes in this form.") is not yet implemented.

**Confirm new email when changing to it**
 
This commit changes the `ChangeUserEmailAddressUseCase` such that the new email address gets immeditely confirmed. This is tested on the use case level by checking that a member that has changed its email address successfully cannot confirm it (because it is already confirmed).

Plan: 371825c0-3112-47cd-9cb0-fea799fad3ee (9x)


4lm (changes requested):
Thanks for adding this feature! I'm not yet done with my review, but I have found a user flow issue I want to discuss before I continue.

If I go to `/user/account` and request an email change, the email link sent to my email account (prompt, in my dev environment) goes to `/user/change-email/...`. There, I can confirm my (in `/user/account`) changed email with a (yes|no) option. If I now hit "Yes," I get a success toast of "Email address changed successfully." I'm totally fine with this user flow.

But, after successfully changing my email, if I go to the dashboard `/member/dashboard,` I get an initial user welcome screen that tells me to confirm my user account. My only option now is to hit "Resent Email" to get an email with a confirmation link to (`/confirm-member/...`) to use my already confirmed account.

I guess this user flow is not intended - right?

sloschert:
> I guess this user flow is not intended - right?

Thanks for raising this. I will try to provide a fix for this.

4lm (approved):
Nice, works for me now. Approved :)

sloschert:
Thanks I merge and register 9x productive consumption in our app collective account.

Issue #1029, opened 2024-06-17
Companies need a list of their registered working hours


sloschert:
Description:

Companies are registering the hours worked of their workers. But currently they cannot see in the app for whom they registered hours worked. For data protection reasons, in the publicly visible account A (`/user/company/{company_id}/account_a`) and in the publicly visible transactions overview (`/user/company/{company_id}/transactions`), only date and amount of the registration is visible.

Proposed solution:

Each company should have a new tile on dashboard section "Company accounting", named "Registered hours worked". There, the company should be able to access all hours they registered, including name and ID of the workers they registered hours for.  

Implementation:

- [x] `ListRegisteredHoursWorkedUseCase`  in `arbeitszeit`
- [x] `ListRegisteredHoursWorkedController` in `arbeitszeit_web`
- [x] `ListRegisteredHoursWorkedPresenter` in `arbeitszeit_web`
- [ ] GET route `list_registered_hours_worked` in `arbeitszeit_flask/company/routes.py`
- [ ] `ListRegisteredHoursWorkedView` in `arbeitszeit_flask/views`
- [ ] Link to route on company dashboard `arbeitszeit_flask/templates/company/dashboard.html`

PR #1034, opened 2024-06-30
Create ListRegisteredHoursWorkedUseCase

sloschert:
The request object for this use case takes a `company_id`. We expect this company to exist in the database.

Pull Requests targeting web and flask layer will follow.

See issue #1029 for motivation.

Plan: 371825c0-3112-47cd-9cb0-fea799fad3ee (2x)

### DIFF
--- a/arbeitszeit/use_cases/list_registered_hours_worked.py
+++ b/arbeitszeit/use_cases/list_registered_hours_worked.py
@@ -1,0 +1,54 @@
+from dataclasses import dataclass
+from datetime import datetime
+from decimal import Decimal
+from uuid import UUID
+
+from arbeitszeit.records import AccountTypes
+from arbeitszeit.repositories import DatabaseGateway
+
+
+@dataclass
+class Request:
+    company_id: UUID
+
+
+@dataclass
+class RegisteredHoursWorked:
+    hours: Decimal
+    worker_id: UUID
+    worker_name: str
+    registered_on: datetime
+
+
+@dataclass
+class Response:
+    registered_hours_worked: list[RegisteredHoursWorked]
+
+
+@dataclass
+class ListRegisteredHoursWorkedUseCase:
+    database_gateway: DatabaseGateway
+
+    def list_registered_hours_worked(self, request: Request) -> Response:
+        company = (
+            self.database_gateway.get_companies().with_id(request.company_id).first()
+        )
+        assert company
+        work_account = company.get_account_by_type(AccountTypes.a)
+        assert work_account
+        records = (
+            self.database_gateway.get_transactions()
+            .where_account_is_sender(work_account)
+            .ordered_by_transaction_date(descending=True)
+            .joined_with_receiver()
+        )
+        registered_hours_worked = [
+            RegisteredHoursWorked(
+                hours=transaction.amount_sent,
+                worker_id=worker.id,
+                worker_name=worker.get_name(),
+                registered_on=transaction.date,
+            )
+            for transaction, worker in records
+        ]
+        return Response(registered_hours_worked=registered_hours_worked)
--- a/arbeitszeit/repositories.py
+++ b/arbeitszeit/repositories.py
@@ -364,6 +364,10 @@ class TransactionResult(QueryResult[records.Transaction], Protocol):
         UUID.
         """
 
+    def joined_with_receiver(
+        self,
+    ) -> QueryResult[Tuple[records.Transaction, records.AccountOwner]]: ...
+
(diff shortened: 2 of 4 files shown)


4lm (approved):
Well done! I approve this PR.

I can only think of little things, probably in the YAGNI category. So, no need to act on them. Nevertheless, I don't want to withhold them from you. Maybe you find some of them useful:

- Why not define ascending/descending in `list_registered_hours_worked` via a `Request` option with descending as a default?
- Maybe adding `transaction.id` to ` RegisteredHoursWorked` could make sense?
- What about limit and offset for listings?
- What about a date range for listings?
- What about listings for a specific worker?

arbeitszeit/use_cases/list_registered_hours_worked.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def list_registered_hours_worked(self, request: Request) -> Response:
+        company = (
+            self.database_gateway.get_companies().with_id(request.company_id).first()
+        )
+        assert company
+        work_account = company.get_account_by_type(AccountTypes.a)
+        assert work_account
+        records = (
+            self.database_gateway.get_transactions()
```
4lm, line 40: Wait a second 🤔, why does the "Use Cases" (Interactor) layer contain a gateway from the outer "Interface Adapters" layer? I think the inner layers don't know the outer layers; the outer layers only operate on the inner layer via Dependency Injection (DI). Please compare with [this diagram](https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html) by Uncle Bob. Or am I missing something? I'm not proficient in Clean Architecture™, so my understanding of it might be wrong.

4lm (changes requested):
I might have found a little roadblock. Therefore, I retract my approval until further discussion. Please look at the code comment in the use case for further info. Thank you!

arbeitszeit/use_cases/list_registered_hours_worked.py
4lm: Also, [the article](https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html) below the linked diagram might be of interest, e.g.:

> [...]
> The Dependency Rule
> [...]
> The overriding rule that makes this architecture work is The Dependency Rule. This rule says that source code dependencies can only point inwards. Nothing in an inner circle can know anything at all about something in an outer circle. In particular, the name of something declared in an outer circle must not be mentioned by the code in the an inner circle. That includes, functions, classes. variables, or any other named software entity.
> [...]
> Use Cases
> [...]
> We do not expect changes in this layer to affect the entities. We also do not expect this layer to be affected by changes to externalities such as the database, the UI, or any of the common frameworks. This layer is isolated from such concerns.
> [...]

sloschert: Hey, 
important point, I hope I can clarify it.

We import the `DatabaseGateway` from the "business logic" module: `from arbeitszeit.repositories import DatabaseGateway`. This DatabaseGateway is an interface (a `Protocol`). We do NOT import the SQL database implementation (`arbeitszeit_flask/database/repositories.py`) nor the in-memory implementation we use for tests (`tests/use_cases/repositories.py`). We inject these implementations when needed.

Then there is still the question: Does these protocol classes know something about SQL database implementations? Some wordings ("join with", "offset", "limit", etc.) seem to point in that direction. But actually I think this abstract interface can be implemented by any kind of repository, be it in-memory, sql, or anything else. That we need some kind of repository in the business logic should be clear.  

Does that make sense and clarify this point?

4lm: Mhh 🤔, I think, I'm in the wrong here! I had a look, in the [Clean Architecture](https://www.oreilly.com/library/view/clean-architecture-a/9780134494272/) book. There in chapter 22 (page 202ff.) on page 208 is a structural (kind of) UML diagram (figure 22.2) which shows that the Use Case Interactor indeed can have a navigable direction towards a Data Access Interface which is comparable to our use of the Database Gateway Protocol (structural sub-typing). Also the text in that chapter seems to confirm this. I will push the state of my review to "approved", once again 😊!

4lm (approved):
Sorry for the back-and-forth. As described in my [code comment](https://github.com/ida-arbeitszeit/arbeitszeitapp/pull/1034#discussion_r1663726909), I re-approve this PR.

sloschert:
@seppeljordan I register 2x consumption

PR #1036, opened 2024-07-03
List registered hours worked (www layer)

sloschert:
**Add ListRegisteredHoursWorkedController** 

This commit adds the `ListRegisteredHoursWorkedController` class, which is responsible for creating a request object for the `list_registered_hours_worked` use case. The controller retrieves the current user role and company from the session and asserts that the current user role is `company`. The company ID is then passed as a parameter to the request object.

See issue #1029 for context.

**Add ListRegisteredHoursWorkedPresenter** 

This commit adds the `ListRegisteredHoursWorkedPresenter` class, which is responsible for presenting the response of the `list_registered_hours_worked` use case. The presenter formats the registered hours worked data, including rounding decimal hours to two decimals and converting worker IDs to strings. This presenter will be used in the web layer to display the registered hours worked information.

See issue #1029 for context.

Plan: 371825c0-3112-47cd-9cb0-fea799fad3ee (4x)

### DIFF
--- a/arbeitszeit_web/www/presenters/list_registered_hours_worked_presenter.py
+++ b/arbeitszeit_web/www/presenters/list_registered_hours_worked_presenter.py
@@ -1,0 +1,38 @@
+from dataclasses import dataclass
+
+from arbeitszeit.use_cases import list_registered_hours_worked
+from arbeitszeit_web.formatters.datetime_formatter import DatetimeFormatter
+
+
+@dataclass
+class RegisteredHoursWorked:
+    hours: str
+    worker_name: str
+    worker_id: str
+    registered_on: str
+
+
+@dataclass
+class ViewModel:
+    registered_hours_worked: list[RegisteredHoursWorked]
+
+
+@dataclass
+class ListRegisteredHoursWorkedPresenter:
+    datetime_formatter: DatetimeFormatter
+
+    def present(self, response: list_registered_hours_worked.Response) -> ViewModel:
+        registered_hours_worked = [
+            RegisteredHoursWorked(
+                hours=str(round(record.hours, 2)),
+                worker_name=record.worker_name,
+                worker_id=str(record.worker_id),
+                registered_on=self.datetime_formatter.format_datetime(
+                    date=record.registered_on,
+                    zone="Europe/Berlin",
+                    fmt="%d.%m.%Y %H:%M",
+                ),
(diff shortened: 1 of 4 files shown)


arbeitszeit_web/www/presenters/list_registered_hours_worked_presenter.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def present(self, response: list_registered_hours_worked.Response) -> ViewModel:
+        registered_hours_worked = [
+            RegisteredHoursWorked(
+                hours=str(round(record.hours, 2)),
+                worker_name=record.worker_name,
+                worker_id=str(record.worker_id),
+                registered_on=self.datetime_formatter.format_datetime(
+                    date=record.registered_on,
+                    zone="Europe/Berlin",
```
4lm, line 32: In another PR, this could be defined via an environment variable. Even better practice is that users should be able to set their specific timezone in their account settings by themselves.

arbeitszeit_web/www/presenters/list_registered_hours_worked_presenter.py
```diff
@@ -0,0 +1,38 @@
...
+                    fmt="%d.%m.%Y %H:%M",
```
4lm, line 33: Same here. For another PR. Formatting could be set via a [language code](https://www.fincher.org/Utilities/CountryLanguageList.shtml) environment variable, like: `de`, `de-DE`, `en`, `en-GB`, `en-US`  of shape: `\<language\>` or `\<language\>-\<country\>`. Even better practice (like above) is that users should be able to set their specific language in their account settings by themselves.

4lm (approved):
I approve this PR 👍 

However, I generally, not specific to this PR, I have some issues with how authentication and authorization are handled in this project (related to scaling and foremost security). I will open an extra issue to discuss this. 

Sorry to be so pedantic, but I think getting those things right is important to avoid future (mostly security) issues that are spotable on the horizon today.

arbeitszeit_web/www/presenters/list_registered_hours_worked_presenter.py
sloschert: Yeah, I agree that this should be configurable..

sloschert:
4xCR

PR #1041, opened 2024-07-14
Fix and improve wording in the mailer environment variable section of the development guide

4lm:
This will be a quick PR review. Just browsed the dev docs and noticed that `SMTP` was incorrectly written as `SMPT`. Fixed this and, while I was at it, improved some minor wording issues in that section as well. Please review - thanks!


sloschert (approved):
Great, thanks!

PR #1027, opened 2024-06-01
Write down some basic considerations for our data model

seppeljordan:
This PR provides some written documentation behind our data model. Please feel free to criticize any aspect of the proposed text. I encourage you to provide a PR for this branch with concrete changes.

Consult the README if you want to render the RST documents from this PR into html.

```
make html
firefox build/html/index.html
```

EDIT: fix bug in instructions

PR #1044, opened 2024-07-16
Implement dedicated recording of hours worked

seppeljordan:
This PR implements the dedicated recording of registered hours worked. Previously we only recorded hours worked via account transaction from the labour account of a company to the individual account of a worker.  This is in fact currently sufficient to reconstruct the hours worked of every single member in the system. This changed was made because it is necessary to move away from the notion that a single registration of hours worked would result in a single transaction in the public ledger but instead would need to split up into several transactions in the future. To elaborate one such transaction for every hour worked would be the portion of labour that is allocated for public plans. In the future we would most certainly want to create a separate transaction that would go into the public sector funds. Other future transactions are to be expected as well if we look towards introducing the notion of accumulation, where an additional portion of every registered work hour would go towards some accumulation funds instead of going into individual consumption.

This commit creates a new table in the DB to serve that end. The tests in the business logic side of things were not subject to change as the behavior of the application should not change in a way observable to the users.

_Plan-ID: 3a003c51-35a3-41a6-9f0a-076d8d7a0975_

### DIFF
--- a/arbeitszeit/records.py
+++ b/arbeitszeit/records.py
@@ -347,3 +347,13 @@ class PasswordResetRequest:
     email_address: str
     reset_token: str
     created_at: datetime
+
+
+@dataclass
+class RegisteredHoursWorked:
+    id: UUID
+    company: UUID
+    member: UUID
+    amount: Decimal
+    transaction: UUID
+    registered_on: datetime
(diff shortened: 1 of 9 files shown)


petsch0:
@seppeljordan in #1027 in the first approach (which is my favorite btw) it was suggested to also record the FIC at that time. Wouldn't it be quite easy to include this in this PR, either within one transaction in the registered_hours_worked table or with the same timestamp in another table "fic"?

petsch0:
that sounds like a reasonable plan @seppeljordan, thanks for clarifying!

sloschert (approved):
Great work. 1xCR.

PR #1045, opened 2024-07-17

seppeljordan:
This commit updates the development dependencies of the arbeitszeitapp.  Several python packages were updated in their version. This updated also includes the updated of the type stubs for the updated python packages.


4lm (commented):
The project [seems to have had some development hick-ups](https://www.reddit.com/r/flask/comments/193huir/why_is_flasklogin_07_not_on_pip/) and flask-login 0.7 was "unreleased," and 0.6.3 seems to be the latest stable and oficial version, which is "Compatibility with Flask 3 and Werkzeug 3," as noted in their [changelog](https://github.com/maxcountryman/flask-login/blob/main/CHANGES.md). 

Can you tell Nix to instead use 0.6.3, to green the checks in this PR? Because as it seems, we could wait for a while until a new version is released. Because people [are waiting for 0.7 since October last year](https://github.com/maxcountryman/flask-login/discussions/798).

4lm (approved):
Looks good to me now (since we have #1058). I therefore approve.

PR #1047, opened 2024-07-17
Provide minified bulma.css version

seppeljordan:
This commit changes the static file
`arbeitszeit_flask/static/bulma.css`. Before we used the regular bulma.css file in our application. With this change we provide instead a "minified" version of the same file. This version of bulma.css which provides the same functionality as the old version is 80 kB smaller. This should provide a significant reduction in network traffic for setups where static files are not compressed for transmission.

@4lm Do you think that this change makes sense?


4lm:
@seppeljordan As I said in the dev meeting, it is not worth the trouble; I'll explain:

Gzipped, both version (normal vs minified CSS file) have only a 3KB difference (68 to 71 KB); being gzipped-delivered is our most relevant and default use case in production. Not being gzipped is primary in developing the case, and there, the connection is super fast, and those 80 KB raw difference (550 vs 630 KB if I recall correctly) between both versions will make in practice, in local development, no difference. But the moment you add a minified version, you now also have to facilitate a [CSS source map](https://web.dev/articles/source-maps) file for debugging purposes (also in production, for better debugging). As I see it, a source map file is provided in the Bulma releases but not linked via a `sourceMappingURL` directive in the distributed minified CSS file. This missing link means we would have to implement now our own build process to create a minified CSS version that links to that CSS map and deliver both files (the source map file is only delivered upon request, thats why it needs to be part of the minified version). Alternatively, you can provide the normal version in development and a minified version in production, which still would need the `sourceMappingURL` to the source map for debugging purposes, hence a custom build process.

Therefore, in my opinion, it's not worth the trouble, also, because we will remove Bulma anyway.

What do you think?

4lm (commented):
@seppeljordan As I said in the dev meeting, it is not worth the trouble; I'll explain:

Gzipped, both version (normal vs minified CSS file) have only a 3KB difference (68 to 71 KB); being gzipped-delivered is our most relevant and default use case in production. Not being gzipped is primary in developing the case, and there, the connection is super fast, and those 80 KB raw difference (550 vs 630 KB if I recall correctly) between both versions will make in practice, in local development, no difference. But the moment you add a minified version, you now also have to facilitate a [CSS source map](https://web.dev/articles/source-maps) file for debugging purposes (also in production, for better debugging). As I see it, a source map file is provided in the Bulma releases but not linked via a sourceMappingURL directive in the distributed minified CSS file. This missing link means we would have to implement now our own build process to create a minified CSS version that links to that CSS map and deliver both files (the source map file is only delivered upon request, thats why it needs to be part of the minified version). Alternatively, you can provide the normal version in development and a minified version in production, which still would need the sourceMappingURL to the source map for debugging purposes, hence a custom build process.

Therefore, in my opinion, it's not worth the trouble, also, because we will remove Bulma anyway.

What do you think?

PR #1048, opened 2024-07-17
Create dedicated flask route for AcceptCooperationRequest use case

seppeljordan:
This PR separates the web logic for the AcceptCooperationRequest use case from the company.my_cooperations routes.  This separation removes the interdepence of the AcceptCooperationRequest web logic with that of the use cases ListCoordinationsOfCompany, DenyCooperation, ListMyCooperatingPlansUseCase and CancelCooperationSolicitation. All existing tests where moved to maintain the same level of test coverage. Also a redimentary test case for the flask routes was implemented.

### DIFF
--- a/arbeitszeit_flask/company/routes.py
+++ b/arbeitszeit_flask/company/routes.py
@@ -5,11 +5,6 @@ from flask import Response as FlaskResponse
 from flask import redirect, render_template, request, url_for
 from flask_login import current_user
 
-from arbeitszeit.use_cases.accept_cooperation import (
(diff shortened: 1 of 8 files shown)


4lm (approved):
Thanks for the PR. Looks good to me 👍 I like how you solved this.

Just one question, which is not a show stopper and just out of interest. Why didn't you use something like `wtforms` (which seems present in the project) and `wtforms.validators.UUID` to construct and validate the form?

sloschert:
1xCR

PR #1049, opened 2024-07-19
Translate latest strings to EN and ES

sloschert:
Parse our code for translatable strings and translate new strings to DE and ES.


sloschert:
Note that the term "Public Sector Fund (PSF)" has been translated as follows:

- DE: Fonds öffentlicher Sektor (FÖS)
- ES: Fondo del Sector Público (FSP)

4lm (approved):
It looks good to me 👍 some line numbers might change after merging my PR in #1039, but if this isn't an immediate problem, go!

sloschert:
These location comments are only intended be a help for the translator, so they can look up the code, it does not effect the translations. 

1xCR

PR #1050, opened 2024-07-20
Return status code 401 on failed login

sloschert:
Before this change we would return 200 when a user (member, company or accountent) provided a wrong password and/or email address on login. Now we return status code 401.


sloschert:
> I think that there should be tests that go with the implementation.

true. have a look now.

4lm (approved):
LGTM. Works as advertised. Approved.

sloschert:
1xCR

PR #1058, opened 2024-07-24
Don't include Flask-Login in contraints.txt

seppeljordan:
Because we use an "unreleased" version of Flask-Login we need to exclude it from constraints.txt. Otherwise users of pip would ne be able to use constraints.txt due to Flask-Login==0.7 not being available on PyPI.


4lm (approved):
Thanks for the PR. LGTM. Also thank you for improving the comments.

PR #1061, opened 2024-07-25
Isolate web logic for DenyCooperation use case

seppeljordan:
Before this commit the web logic for the DenyCooperation was tied up with the `my_cooperations` route. This was undesirable and thusly the web logic was isolated into its own route, view and presenter. The HTML form on the `my_cooperations` route needed to be adjusted slightly. Also a new very basic test for the new route was implemented that checks that the web server at least does not crash when submitting a post request against it.

### DIFF
--- a/arbeitszeit_web/www/presenters/deny_cooperation_presenter.py
+++ b/arbeitszeit_web/www/presenters/deny_cooperation_presenter.py
@@ -1,0 +1,56 @@
+from dataclasses import dataclass
+
+from arbeitszeit.use_cases.deny_cooperation import DenyCooperationResponse
+from arbeitszeit_web.notification import Notifier
+from arbeitszeit_web.translator import Translator
+
+
+@dataclass
+class ViewModel:
+    redirection_url: str
+
+
+@dataclass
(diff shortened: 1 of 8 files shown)


4lm (approved):
Thanks for this complementary PR to your previous PR #1048. LGTM.

arbeitszeit_web/www/presenters/deny_cooperation_presenter.py
```diff
@@ -0,0 +1,56 @@
...
+                        "You are not coordinator of this cooperation."
+                    )
+                )
+            else:
+                # catchall for rejected responses where rejection
+                # reason cannot be handled by presenter.
+                self.notifier.display_warning(
+                    self.translator.gettext("Could not deny cooperation")
+                )
+        return ViewModel(redirection_url="/company/my_cooperations")
```
sloschert, line 56: i notice you didn't use the url_index. Is it because you are not testing the value of `redirection_url`?

sloschert (approved):
1xCR.
I only have one question out of curiosity.

PR #1063, opened 2024-07-27
Add integration test of icon path

sloschert:
This commit adds an integration test that checks if the expected SVG path of the "create new plan icon", as specified in `arbeitszeit_flask/templates/icons/file-circle-plus.html`, appears  on the company dashboard.

I am aware that this test is very specific, but in no other place we test for the concrete icon path value.


sloschert:
thanks for spotting the wrong string. have a look now.

4lm (commented):
Isn't this similiar to the one integration test that cam with my PR? [There](https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/master/tests/flask_integration/test_filters.py#L155-L163) I added an integration test on the "key" icon in the login form, which test the whole svg with path element in the html response.

sloschert:
@4lm You are correct, somehow I did overlook this test. Considering this, perhaps it's best to delete my PR.

PR #1065, opened 2024-07-27
Redirect to my_cooperations view after accept/deny coop request

seppeljordan:
Before this commit we would have the redirection url after accepting or denying a cooperation request by another company hard-coded into the presenter. With this commit we instead defer to UrlIndex to find the correct url for this redirection. There are also tests implemented to check for this fact on the presenter level.


sloschert (approved):
Top, thanks

PR #1068, opened 2024-07-27
Isolate cancel_cooperation_request view

seppeljordan:
Before this commit the functionality to cancel a cooperation request was all tied up with the logic to present the user an overview over their current cooperations and requests. This commit isolates the CancelCooperationSolicitation use case logic into its own flask route.


sloschert (approved):
Thanks for cleaning up. 1xCR

PR #1071, opened 2024-08-06
Unify `ListInboundCoopRequests` and `ListOutboundCoopRequests`

seppeljordan:
This PR unifies the use cases `ListInboundCoopRequests` and `ListOutboundCoopRequests` into one single use case `ShowCompanyCooperationsUseCase`. We did this because the two original use cases were identified be part of a single use case where the users would want do display all the cooperation info for a given company in an overview. The tests and business logic were left intact while being reorganized into unified business logic and test modules.

This is only a step into unifying all the involved use cases around the company cooperation overview. The `ListCoordinationsOfCompany` and `ListMyCooperatingPlansUseCase` classes remain to be unified with the newly created `ShowCompanyCooperationsUseCase` use case class.


arbeitszeit_flask/company/routes.py
```diff
@@ -273,27 +269,22 @@ class request_cooperation(RequestCooperationView): ...
...
 ):
     list_coord_response = list_coordinations(
         ListCoordinationsOfCompanyRequest(UUID(current_user.id))
     )
-    list_inbound_coop_requests_response = list_inbound_coop_requests(
-        ListInboundCoopRequestsRequest(UUID(current_user.id))
-    )
-    list_outbound_coop_requests_response = list_outbound_coop_requests(
-        ListOutboundCoopRequestsRequest(UUID(current_user.id))
+    use_case_response = show_company_cooperations.show_company_cooperations(
```
4lm, line 279: Minor. The wording `use_case_response` seems a little generic the other to response are use case responses as well.

4lm (approved):
Works as advertised. I have one minor thing (see comment) but no blocker. Approved.

PR #1073, opened 2024-08-11
Fix bug in CreateDraftView

seppeljordan:
Before this change there was a bug in CreateDraftView. The was long undetected as it is hard to trigger the failure resulting from the bug. The user would need to submit form data to the view that contained erroneous planning data. This would have led to an exception being raised in a CreatePlanDraft use case object. The application would have "crashed" with a 500 error response instead of rendering the original form back to the user with an error message. The bug was addressed by including proper form validation in POST handler of the method and returning the rendered form to the user when form validation fails.

### DIFF
--- a/arbeitszeit_flask/views/create_draft_view.py
+++ b/arbeitszeit_flask/views/create_draft_view.py
@@ -27,6 +27,8 @@ class CreateDraftView:
     @commit_changes
     def POST(self) -> Response:
         form = CreateDraftForm(request.form)
+        if not form.validate():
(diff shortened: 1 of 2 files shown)


4lm (approved):
I could not provoke a 500 in the view (outside of this PR-branch, of course). Therefore I couldn't reconstruct your initial claim, but it in the end it doesn't matter. Your PR closes the gap towards a default implementation on how to handle a form submit control flow in a web application relying on request response cycles. Therefore I support this PR. Approved.

PR #1074, opened 2024-08-11
Remove NixOS 23.11 from flake inputs

seppeljordan:
This commit removes the definitions for NixOS 23.11 from the flake inputs of this repository.  This was done to reduce build and test times. Since all systems that (are currently known to) run the arbeitszeitapp are on NixOS 24.05 there is no practical need for testing with the older stable version of NixOS.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -55,22 +55,6 @@
         "type": "github"
       }
     },
-    "nixos-23-11": {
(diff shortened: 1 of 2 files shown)


4lm (approved):
OK. LGTM. Approved.

PR #1075, opened 2024-08-11
Changer error propagation in `icon` template filter

seppeljordan:
This change concerns the error propagation inside the `icon` template filter.  Before this change some exception raise as part of the `icon_filter` routine were changed to include the icon name as part of the exception message.  After this change the message of those exceptions will be left as is and instead the `add_note` method is used to add further context to the exception. The `add_note` method is new in Python 3.11. See
https://docs.python.org/3.11/library/exceptions.html#BaseException.add_note for more information on the topic.

Also the error message slightly changed to indicate that exceptions occurred during the loading process of a specific icon.

/cc @4lm 

### DIFF
--- a/arbeitszeit_flask/filters.py
+++ b/arbeitszeit_flask/filters.py
@@ -67,38 +67,36 @@ def icon_filter(
     section in the developement guide.
     """
     try:
-        # Treat empty icon_name as intentionally set null value
-        if not icon_name.strip():
-            return Markup("")
-
-        file_name = f"{icon_name}.html"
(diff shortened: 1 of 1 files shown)


4lm (approved):
Nice one! My stomach actually felt a little weird when implemented the exception text enrichment by wrapping the original exception with some custom text. Having an idiomatic function call in 3.11 to solve this feels nice. Therefore, approved. PS: I don't have strong feelings concerning extracting the logic into a second function call. Both reads IMO equally good (bad). So I go along with your change in this regard.

PR #1076, opened 2024-08-12
Implement PreviewPrivateConsumptionUseCase

seppeljordan:
This PreviewPrivateConsumptionUseCase implemented in this commit is supposed to be included in the workflow for members to register their individual consumption of goods and services through the app.  The use case is not yet hooked into the web code of the app. If hooked into the web code this use case would provide basic product information for a given plan id and the total consumption cost in case the amount is also provided. The information will be displayed for the user so that they can more easily identify what kind of product and how much they are about to "consume".

### DIFF
--- a/arbeitszeit/use_cases/preview_private_consumption.py
+++ b/arbeitszeit/use_cases/preview_private_consumption.py
@@ -1,0 +1,46 @@
+from dataclasses import dataclass
+from decimal import Decimal
+from uuid import UUID
+
+from arbeitszeit.repositories import DatabaseGateway
+
+
+@dataclass
+class Request:
+    plan: UUID | None = None
+    amount: int | None = None
+
+
+@dataclass
+class ProductInformation:
+    name: str | None = None
+    description: str | None = None
+
+
+@dataclass
+class Response:
+    product: ProductInformation | None = None
+    total_cost: Decimal | None = None
+
+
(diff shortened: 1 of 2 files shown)


4lm (approved):
Thanks for providing this PR. I have some remarks, which I think may boil down to personal taste. Therefore I approve this PR but have some remarks concerning the call function of `PreviewPrivateConsumptionUseCase`.

In my view, the control flow within the `preview_private_consumption` call function feels somewhat busy, making it less unclear than needed regarding the what, when, and why of its operations. Additionally, assigning `None` values seems redundant since these are already the default values for the relevant data classes. The function also continues processing even after it has gathered all the necessary information (because it only has one return at the end), which could be streamlined to enhance readability and improve performance. I would recommend refactoring `preview_private_consumption` to something along the lines of the following:

```python
def preview_private_consumption(self, request: Request) -> Response:
        if not request.plan:
            return Response()

        plan = self.database.get_plans().with_id(request.plan).first()

        if not plan:
            return Response()

        product = ProductInformation(
            name=plan.prd_name,
            description=plan.description,
        )

        if request.amount is None or request.amount <= 0:
            return Response(product=product)

        total_cost = plan.production_costs.total_cost() * request.amount

        return Response(
            product=product,
            total_cost=total_cost,
        )
```

PR #1077, opened 2024-08-12

seppeljordan:
This update was produced by running

```bash
python -m arbeitszeit_development.update_dependencies && python format_code.py
```


4lm (approved):
Looks good and update ran as advertised. Approved.

PR #1078, opened 2024-09-08
Implement selection of productive consumption

### LINKED ISSUE #1053: RFC: Change productive consumption user flow
I propose to change our productive consumption user flow, please let me know what you think about this.

Current user flow
==================

For registering of productive consumption we currently have one route, where companies are registering productive consumption: `company/register_productive_consumption`. This route accepts GET and POST requests. The POST request requires three inputs: `plan_id`, `amount` and `type_of_consumption`. Companies have to make sure to specify the correct plan_id, because at no point of this flow the plan name or the "price" of the product is shown, which is error prone.

Proposed user flow
==================

I propose to split the process of registering productive consumption into two stages:

STAGE 1: Selection of plan
--------------------------- 

In this stage, the company specifies the plan_id of the product that it wants to consume.

We can have different implementations of this stage. For example, the selection of the plan_id might be carried out by scanning a plan-specific QR-Code, by clicking a link somewhere in our web app or by manually entering the plan_id in a form. For now, I propose to add a new route, that lets the company specify the plan_id (see below for details). 

If the selected plan is eligible, the company gets directed to stage two.

STAGE 2: Selection of amount and type of consumption
----------------------------------------------------

The user gets provided with information on the previously selected plan/product. This information might be the "price" of the product, name and description, and potentially many more. 

The company then specifies amount and type_of_consumption and actually registers the consumption.

Technical implementation:
==========================

STAGE 1
--------

- New company route `/company/select_plan_for_consumption` [GET, POST]
- New view `SelectPlanForConsumptionView`
- New form `SelectPlanForConsumptionForm`

GET:
- New form is shown with one field ("plan_id")

POST:
- New controller `SelectPlanForConsumptionController` 
- New `SelectPlanForConsumptionUseCase`
	- Check that plan exists, that it's active, that it's not public, that consumer is not planner, etc.  
- New presenter `SelectPlanForConsumptionPresenter`
	- On success, redirect to route `/company/register_productive_consumption` (stage 2)
	- On failure, show error notification to user.
	
STAGE 2
--------

- Change existing route `/company/register_productive_consumption` [GET, POST]
- Change existing view `RegisterProductiveConsumptionView`
- Change existing form `RegisterProductiveConsumptionForm`

GET:
- New `ProductiveConsumptionDetailsController`
	- Accept plan_id as argument
	- Handle form
- New `ProductiveConsumptionDetailsUseCase`
	- get information of selected plan from repositories 
- New `ProductiveConsumptionDetailsPresenter`
	
POST:
- Change existing controller `RegisterProductiveConsumptionController`
- Change existing use case `RegisterProductiveConsumption`
	- Check sufficient account coverage etc.
- Change existing presenter `RegisterProductiveConsumptionPresenter`

sloschert:
This commit creates SelectProductiveConsumptionUseCase, -Presenter and -Controller and uses these in the RegisterProductiveConsumptionView. They are intended to handle the GET part of this view, leaving POST mostly untouched.

It implements a user flow described in issue #1053. New features:

* The user sees some plan info before they have the chance to register consumption.
* The provided plan ID appears as a read-only form field if it is valid.
* There are two new buttons ("Select" and "Cancel").

That is IMO quite a lot of (quite complex) code for a not-so-huge change. Until now it took me around 25 hours (including failed attempts) and I'm still not really happy with the outcome. Perhaps we can discuss in one of our next meetings if this is a necessary consequence of clean architecture (which might be true) or if we can make things easier. I imagine that implementing something like this is even more frustrating for newcomers. 

fixes #1053

Plan: 8ec03a5b-3dbe-465d-a533-4b3bfe16121b (26x)

### DIFF
--- a/arbeitszeit/use_cases/select_productive_consumption.py
+++ b/arbeitszeit/use_cases/select_productive_consumption.py
@@ -1,0 +1,70 @@
+from dataclasses import dataclass
+from uuid import UUID
+
+from arbeitszeit.datetime_service import DatetimeService
+from arbeitszeit.records import ConsumptionType
+from arbeitszeit.repositories import DatabaseGateway
+
+
+@dataclass
+class Request:
+    plan_id: UUID | None
+    amount: int | None
+    consumption_type: ConsumptionType | None
+
+
+@dataclass
+class NoPlanResponse:
+    amount: int | None
+    consumption_type: ConsumptionType | None
+
+
+@dataclass
+class InvalidPlanResponse:
+    amount: int | None
+    consumption_type: ConsumptionType | None
+
+
+@dataclass
+class ValidPlanResponse:
+    plan_id: UUID
+    amount: int | None
+    consumption_type: ConsumptionType | None
+    plan_name: str
+    plan_description: str
+
+
+@dataclass
+class SelectProductiveConsumptionUseCase:
+    database: DatabaseGateway
+    datetime_service: DatetimeService
+
+    def select_productive_consumption(
+        self, request: Request
+    ) -> NoPlanResponse | InvalidPlanResponse | ValidPlanResponse:
+        amount = request.amount
+        consumption_type = request.consumption_type
+        if not request.plan_id:
+            return NoPlanResponse(
+                amount=amount,
+                consumption_type=consumption_type,
+            )
+        plan_result = (
+            self.database.get_plans()
+            .with_id(request.plan_id)
+            .that_will_expire_after(self.datetime_service.now())
+        )
+        if not plan_result:
+            return InvalidPlanResponse(
+                amount=amount,
+                consumption_type=consumption_type,
+            )
+        plan = plan_result.first()
+        assert plan
+        return ValidPlanResponse(
+            plan_id=request.plan_id,
+            amount=amount,
+            consumption_type=consumption_type,
+            plan_name=plan.prd_name,
+            plan_description=plan.description,
+        )
--- a/arbeitszeit_flask/templates/company/register_productive_consumption.html
+++ b/arbeitszeit_flask/templates/company/register_productive_consumption.html
@@ -21,14 +21,35 @@
             </div>
             {% endfor %}
             {% endfor %}
+            {% if view_model.valid_plan_selected is defined and view_model.valid_plan_selected %}
             <form method="post">
                 <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
+            {% else %}
+            <form method="get">
+            {% endif %}
                 <div class="field">
                     <label class="label">{{ gettext("Plan ID") }}</label>
                     <div class="control">
+                        {% if view_model.valid_plan_selected is defined and view_model.valid_plan_selected %}
+                        {{ form.plan_id(class_="input", readonly=true) }}
+                        {% else %}
                         {{ form.plan_id(class_="input") }}
+                        {% endif %}
                     </div>
                 </div>
+                {% if view_model.valid_plan_selected is defined and view_model.valid_plan_selected %}
+                <div class="notification is-info">
+                    <p>
+                        {{ gettext("Plan ID") }}: {{ view_model.plan_id }}
+                    </p>
+                    <p>
+                        {{ gettext("Plan name") }}: {{ view_model.plan_name }}
+                    </p>
+                    <p>
+                        {{ gettext("Plan description") }}: {{ view_model.plan_description }}
+                    </p>
+                </div>
+                {% endif %}
                 <div class="field">
                     <label class="label">{{ gettext("Amount") }}</label>
                     <div class="control">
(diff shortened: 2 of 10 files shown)


arbeitszeit/use_cases/select_productive_consumption.py
```diff
@@ -0,0 +1,70 @@
...
+    amount: int | None
+    consumption_type: ConsumptionType | None
+
+
+@dataclass
+class InvalidPlanResponse:
+    amount: int | None
+    consumption_type: ConsumptionType | None
+
+
```
Amittai-Aviram, line 27: All these classes have the same structure.  Wouldn't this be an obvious case for inheritance?

Amittai-Aviram: It is a software engineering benefit:  it avoids code duplication.  If you wish later to change the basic pattern — say, to add a new field — it is easy to do that in one place.  So it makes your code cleaner.

Amittai-Aviram (commented):
Looks good so far!  This is a fairly big PR, so I'm afraid it is taking me some time to get through it, and I am reviewing in pieces.

arbeitszeit_flask/templates/company/register_productive_consumption.html
```diff
@@ -21,14 +21,35 @@ <h1 class="title">
...
+                {% if view_model.valid_plan_selected is defined and view_model.valid_plan_selected %}
+                <div class="notification is-info">
+                    <p>
+                        {{ gettext("Plan ID") }}: {{ view_model.plan_id }}
+                    </p>
+                    <p>
+                        {{ gettext("Plan name") }}: {{ view_model.plan_name }}
+                    </p>
+                    <p>
+                        {{ gettext("Plan description") }}: {{ view_model.plan_description }}
```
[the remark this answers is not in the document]
sloschert: When a company files a plan, it has to provide, besides other things, a plan name and a plan description. That's where this description comes from.

arbeitszeit_flask/templates/company/register_productive_consumption.html
```diff
@@ -21,14 +21,35 @@ <h1 class="title">
...
                     <label class="label">{{ gettext("Plan ID") }}</label>
                     <div class="control">
+                        {% if view_model.valid_plan_selected is defined and view_model.valid_plan_selected %}
+                        {{ form.plan_id(class_="input", readonly=true) }}
+                        {% else %}
                         {{ form.plan_id(class_="input") }}
+                        {% endif %}
                     </div>
                 </div>
+                {% if view_model.valid_plan_selected is defined and view_model.valid_plan_selected %}
```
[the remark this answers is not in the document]
sloschert: Not that I know of... We do the "is defined" check here, because the view_model variable may not be defined at all.

arbeitszeit_flask/templates/company/register_productive_consumption.html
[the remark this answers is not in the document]
Amittai-Aviram: So if it's the same as the description that the producer had entered together with the Plan ID, why should the user even enter it at all?  I do not understand the logic here.  I am sorry if this is obvious to everybody else except me.

arbeitszeit_web/www/presenters/select_productive_consumption_presenter.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @dataclass
+    class ViewModel:
+        valid_plan_selected: bool
+        plan_id: str | None
+        plan_name: str | None
+        plan_description: str | None
+        amount: int | None
+        is_consumption_of_fixed_means: bool
+        status_code: int
```
Amittai-Aviram, line 26: From my POV, ideally, the type of **status_code** would be an enumeration type, where the names of the enumerators would be the standard meanings (OK, NOT_FOUND, etc.), and the values would be there standard values.  The benefit of this is clearer self-documenting code.  We should actually define such an enumeration type so that it is globally available.  But I can understand why you might not wish to bother.

arbeitszeit_web/www/presenters/select_productive_consumption_presenter.py
```diff
@@ -0,0 +1,68 @@
...
+            )
+            return self.ViewModel(
+                valid_plan_selected=False,
+                plan_id=None,
+                plan_name=None,
+                plan_description=None,
+                amount=response.amount,
+                is_consumption_of_fixed_means=response.consumption_type
+                == ConsumptionType.means_of_prod,
+                status_code=404,
```
Amittai-Aviram, line 56: But the user does not actually get a 404 response, does he/she?  Don't we redirect to the form with a meaningful error message?

Amittai-Aviram (approved):
LGTM.  Just please have a look at my new comments, if you would.  Of course, you can always ignore them.  🙂

arbeitszeit_web/www/presenters/select_productive_consumption_presenter.py
sloschert: They do indeed get a 404, but also a meaningful error message ("The selected plan does not exist or is not active anymore.")
Would a 400 be preferable?

arbeitszeit_web/www/presenters/select_productive_consumption_presenter.py
sloschert: I prefer not to add to this already big PR a structural change. But I agree that this is not ideal as it is.

arbeitszeit_web/www/presenters/select_productive_consumption_presenter.py
sloschert: I actually just copied this behaviout from the `RegisterPrivateConsumptionPresenter`. 

I'll merge now as it is, we can change it always afterwards.

sloschert:
@seppeljordan Please register consumption...

PR #1083, opened 2024-09-10
Reimplement register_private_consumption web logic

seppeljordan:
This commit reimplements the web logic for registering a private consumption in the arbeitszeitapp. Previously our approach relied heavily on wtforms. This approarch proved inflexible since it is very hard to manipulate field properties of wtforms forms reliably and safely. Hence, we reimplemented the web logic with standard templating techniques where the forms data in encoded in a basic "dataclass".

This change is made in preparation for an upcoming change where the individual fields of the form in question can be set to "readonly" conditionally.

### DIFF
--- a/arbeitszeit_flask/views/register_private_consumption.py
+++ b/arbeitszeit_flask/views/register_private_consumption.py
@@ -1,5 +1,4 @@
 from dataclasses import dataclass
-from typing import Optional
 
 from flask import Response as FlaskResponse
 from flask import redirect, render_template, request, url_for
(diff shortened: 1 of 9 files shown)


arbeitszeit_flask/views/register_private_consumption.py
```diff
@@ -28,38 +28,34 @@ class RegisterPrivateConsumptionView:
...
+            use_case_request = self.controller.import_form_data(
+                current_user, FlaskRequest()
+            )
+        except self.controller.FormError as error:
+            return FlaskResponse(self._render_template(error.form), status=400)
         response = self.register_private_consumption.register_private_consumption(
             use_case_request
         )
-        view_model = self.presenter.present(response)
+        view_model = self.presenter.present(response, request=FlaskRequest())
```
sloschert, line 50: same question here (the request instance is not used in the presenter, btw)

arbeitszeit_web/forms.py
```diff
@@ -89,3 +84,11 @@ def cooperation_field(self) -> FormField[str]: ...
 
 class ConfirmEmailAddressChangeForm(Protocol):
     def is_accepted_field(self) -> FormField[bool]: ...
+
+
+@dataclass
+class RegisterPrivateConsumptionForm:
+    plan_id_value: str
```
sloschert, line 91: Why do you think it's necessary to add `_value`?

sloschert (changes requested):

sloschert:
Ok thanks for the explanations. 1xCR

sloschert (approved):

Issue #1084, opened 2024-09-10
Remove use of `@commit_changes` in GET request methods


4lm:
I started reviewing #1078, which may take a while to complete. There, I found an issue that is also present in other places in the source code, and which I therefore want to discuss in this RFC and not directly in the PR because I think it would deviate too much from the PR review. I have a question concerning the use of `@commit_changes` on GET request methods. I see this property is also used in other GET request view methods.

Example: https://github.com/ida-arbeitszeit/arbeitszeitapp/pull/1078/files#diff-cd0feb128dcca882b140b71c5409ef24ac8ecda6d80f137ea9e87ecf4f83bbfbR38

In my understanding, this property calls `db.session.commit()` after the wrapped function executes. If true, I don't understand why this is used in GET requests because the result of calling a resource with a GET request should never create, update or delete any database entries. GET requests should only be used to retrieve information. Only on POST, PATCH, PUT, and DELETE request methods does the use of `db.session.commit()` make sense, right? Or am I missing something here? Calling `db.session.commit()` would only make sense if we log user behavior straight to our database (tracking), which we don't do - right?

Please compare with the MDN documentation concerning GET requests:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET
- https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP

Money quote:

> [...] a method is safe if it leads to a read-only operation. Several common HTTP methods are safe: GET, HEAD, or OPTIONS. All safe methods are also idempotent [...] Even if safe methods have a read-only semantic, servers can alter their state: e.g. they can log or keep statistics. What is important here is that by calling a safe method, the client doesn't request any server change itself, and therefore won't create an unnecessary load or burden for the server. Browsers can call safe methods without fearing to cause any harm to the server; this allows them to perform activities like pre-fetching without risk.

sloschert:
In theory, we should just be able to delete the `@commit_changes` decorator from GET request methods, because I agree that it should not be there. In practice, I would check the code before doing so, so that we do not accidentally delete some (misplaced) functionality.

sloschert:
Problematic decorators should get deleted from the following places:

- [ ] `EndCooperationView` in `arbeitszeit_flask/views/end_cooperation_view.py`
- [ ] `company/my_cooperations` route
- [ ] `RegisterHoursWorkedView` in `arbeitszeit_flask/views/register_hours_worked_view.py`
[...]

PR #1086, opened 2024-09-11
Delete @commit_changes decorator from GET routes

sloschert:
This commit addresses nearly all GET routes listed in issue #1084, only `EndCooperationView` in `arbeitszeit_flask/views/end_cooperation_view.py` needs some more thorough refactoring.

Plan: 8ec03a5b-3dbe-465d-a533-4b3bfe16121b (1x)

### DIFF
--- a/arbeitszeit_flask/views/request_coordination_transfer_view.py
+++ b/arbeitszeit_flask/views/request_coordination_transfer_view.py
@@ -24,7 +24,6 @@ class RequestCoordinationTransferView:
     controller: RequestCoordinationTransferController
     use_case: RequestCoordinationTransferUseCase
 
-    @commit_changes
     def GET(self, coop_id: UUID) -> types.Response:
         form = RequestCoordinationTransferForm()
         form.cooperation_field().set_value(str(coop_id))
--- a/arbeitszeit_flask/company/routes.py
+++ b/arbeitszeit_flask/company/routes.py
@@ -266,7 +266,6 @@ class request_cooperation(RequestCooperationView): ...
 
 
 @CompanyRoute("/my_cooperations", methods=["GET"])
-@commit_changes
 def my_cooperations(
     list_coordinations: ListCoordinationsOfCompany,
     show_company_cooperations: ShowCompanyCooperationsUseCase,
--- a/arbeitszeit_flask/views/register_hours_worked_view.py
+++ b/arbeitszeit_flask/views/register_hours_worked_view.py
@@ -23,7 +23,6 @@ class RegisterHoursWorkedView:
     presenter: RegisterHoursWorkedPresenter
     list_workers: ListWorkers
 
-    @commit_changes
     def GET(self) -> Response:
         return self.create_response(status=200)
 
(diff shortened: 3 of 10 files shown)


sloschert:
> This change would break the "End" button at the bottom of the plan details page. This button is rendered only if a plan is viewed by its planning company and the plan is part of a cooperation. The button is currently a `<a>` element and would need to change to being a `<form>` element.

Thanks for catching. The same is true for the "end cooperation" button on coop summary pages. I'm on it.

arbeitszeit_flask/views/request_coordination_transfer_view.py
```diff
@@ -24,7 +24,6 @@ class RequestCoordinationTransferView:
     controller: RequestCoordinationTransferController
     use_case: RequestCoordinationTransferUseCase
 
-    @commit_changes
```
Amittai-Aviram, line 27: I think there might be another problem here with templates/user/coop_summary.html" — another link in the form of a button for transferring the coordination.  It's line 37 in master.

Amittai-Aviram (commented):
This is an excellent change!  But we need to look through the code base to find links serving as triggers for state changes.  I found one — maybe that is the only one left.

arbeitszeit_flask/views/request_coordination_transfer_view.py
sloschert: You mean this, right?

```python
{% if view_model.transfer_coordination_url is not none %}
      <a class="button is-small is-primary" href="{{ view_model.transfer_coordination_url }}">
          <span class="icon">
              {{ "square-arrow-up-right"|icon }}
          </span>
          <span>{{ gettext("Transfer coordination") }}</span>
      </a>
 {% endif %}
  ```
  
  This link does lead to the `request_coordination_transfer` route in the company blueprint. But it leads to the GET method of this view and there is no state change implied, so there should be no problem.

Comment (sloschert) [reply arbeitszeit_flask/views/request_coordination_transfer_view.py:27]: Let me know if this can get merged from your point of view.

**Review by Amittai-Aviram — APPROVED** [review]
LGTM.  Thanks for your patience.

File: arbeitszeit_flask/templates/user/coop_summary.html [hunk 2]
```diff
@@ -78,12 +78,17 @@ <h1 class="title is-4 has-text-centered">
                     </div>
                     <div>
                         {% if plan.show_end_coop_button %}
-                        <a class="button is-small is-danger" href="{{ plan.end_coop_url }}">
-                            <span class="icon">
-                                {{ "times"|icon }}
-                            </span>
-                            <span>{{ gettext("End cooperation") }}</span>
-                        </a>
+                         <form action="{{ url_for('main_company.end_cooperation') }}" method="post">
```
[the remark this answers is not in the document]
sloschert: If I recall correctly it was probably the (wrong) assumption, that using the url_index was less helpfull if there is only a POST request. If you fin it important for me to change it let me know here and I will open a ticket to change it later.

PR #1087, opened 2024-09-15
Move redirection decision into register_private_consumption presenter

seppeljordan:
This commit moves the responsibility for a possible redirect into the presenter when dealing with forms for registering private consumptions.


sloschert (approved):
Top. 1x CR

PR #1089, opened 2024-09-18
Show proper form errors when registering private consumption fails

seppeljordan:
Before this commit we would only show problems with registering a private consumption in a errors notification that would flash briefly on the screen of the user. This commit changes the behavior such that errors messages are diplayed "permanently" on the users screen, i.e. those errors messages don't fade out.


sloschert (approved):
:+1:  1xCR

PR #1097, opened 2024-09-24
Redirect unautherized users for register_private_consumption

seppeljordan:
The commit implements controller based redirection logic for unautherized users on the level of the controller. The RegisterPrivateConsumptionController can now return a `Redirect` object to direct users to the login page when they try to register a private consumption while not being logged in.


sloschert (approved):
Thanks, 1x CR

PR #1102, opened 2024-10-28
Translation of latest strings to DE and ES 

sloschert:


arbeitszeit_flask/translations/de/LC_MESSAGES/messages.po
```diff
@@ -852,6 +863,18 @@ msgstr "Noch keine Arbeiter*innen registriert"
 msgid "Registration of productive consumption"
 msgstr "Registrierung produktiven Konsums"
 
+#: arbeitszeit_flask/templates/company/register_productive_consumption.html:46
+msgid "Plan name"
+msgstr "Planname"
```
petsch0, line 868: those put-together nouns look strange in german imo. Can we not get rid of them at all, and simply use "Name-Name" and "description-Beschreibung"? This maybe could be re-used at other places as well. But of course, it would need to be clear from the context that, e.g. name of a PLAN is referred to.

arbeitszeit_flask/translations/de/LC_MESSAGES/messages.po
```diff
@@ -852,6 +863,18 @@ msgstr "Noch keine Arbeiter*innen registriert"
...
+
+#: arbeitszeit_flask/templates/company/register_productive_consumption.html:49
+msgid "Plan description"
+msgstr "Planbeschreibung"
```
petsch0, line 872: dito

arbeitszeit_flask/translations/de/LC_MESSAGES/messages.po
```diff
@@ -1700,26 +1734,38 @@ msgstr "Ungültige Plan-ID."
 msgid "Invalid cooperation ID."
 msgstr "Ungültige Kooperations-ID."
 
-#: arbeitszeit_web/www/presenters/accept_cooperation_request_presenter.py:21
+#: arbeitszeit_web/www/controllers/select_productive_consumption_controller.py:46
+msgid "The provided plan ID is not a valid UUID."
+msgstr "Die angegebene Plan-ID ist keine gültige UUID."
+
+#: arbeitszeit_web/www/controllers/select_productive_consumption_controller.py:66
+msgid "The provided amount is not a valid integer."
```
petsch0, line 1742: from the german translation, shouldn't this be "The provided amount is not a valid number."?

arbeitszeit_flask/translations/de/LC_MESSAGES/messages.po
```diff
@@ -2124,6 +2198,10 @@ msgstr "Unerwarteter Fehler: Nicht möglich, die Planeinreichung zu wiederrufen.
 msgid "Filing of plan has been successfully revoked."
 msgstr "Planeinreichung wurde erfolgreich widerrufen."
 
+#: arbeitszeit_web/www/presenters/select_productive_consumption_presenter.py:44
+msgid "The selected plan does not exist or is not active anymore."
+msgstr "Der ausgewählte Plan existiert nicht oder ist nicht mehr aktiv."
```
petsch0, line 2203: how can this happen? A plan can be selected, but does not exist?

arbeitszeit_flask/translations/de/LC_MESSAGES/messages.po
sloschert: see comment below

arbeitszeit_flask/translations/de/LC_MESSAGES/messages.po
sloschert: I think the correct translation for "ganze Zahl" is "integer", and vice versa. https://de.pons.com/%C3%BCbersetzung-2/englisch-deutsch/integer

arbeitszeit_flask/translations/de/LC_MESSAGES/messages.po
sloschert: See this discussion (https://github.com/ida-arbeitszeit/arbeitszeitapp/issues/1053) and the screenshot below. We agreed on having a two-stage productive consumption process: First the user "selects" a plan and if this plan exists more details are shown. So it may happen that a user selects a plan that does not exist.

alt text

petsch0: I see, thanks! so the plan id is not really selected, but instead it is a free-text field. In this case, sth like "a plan with this plan id does not exist" would be clearer. but such a minor issue can be taken up later imo.

petsch0 (approved):
lgtm, thanks @sloschert!

arbeitszeit_flask/translations/de/LC_MESSAGES/messages.po
sloschert: yes, you are right.

Issue #1092, opened 2024-09-20
RFC Delete worker from company


sloschert:
Currently it is not possible to undo a work relationship between worker and company.

I propose that companies should be able to delete workers from their list of associated workers. 

On company route `/company/invite_worker_to_company` there is currently a list of associated workers displayed. I propose to add a "delete" button to each worker entry.

After clicking that button and confirming a warning, the worker should be deleted from the list of associated workers and an informational email should be sent to the worker. 
 
This would be a minimal implementation, making it only possible for the company to delete workers from the list of associated workers and intentionally does not include any similar feature on the worker side.  

For simplicity and prioritization, I propose not to keep any history of worker's past workplaces or the company's past workers in the database for now. Instead, simply a row in the `jobs` table would get deleted.

What do you think?

Amittai-Aviram:
I agree that it is a good starting point.  My main concern is the route:  the user interface for deleting an employee is in the page sent back from a request to `/company/invite_worker_to_company`.  This seems to me to be an odd place in which to have this feature.  I guess the manager does not necessarily have to invite a new employee, he or she only needs to request that page.  For now, perhaps this is OK.

sloschert:
Yes, I agree that this is a odd place. Perhaps some changes to this route are necessary in the process.

(But yes, it is possible to access this page via GET, to receive the list of employees without inviting an employee.)

sloschert:
FYI
My current plan is to create a new route (accepting GET and POST requests), called `company/select_worker_to_remove_from_company`.

The view class of this route implements two use cases (together with their controllers, presenter, etc.) of which the first one is still completely missing:

GET:

- SelectWorkerToRemoveFromCompany

POST:

- RemoveWorkerFromCompany

The integration with the current code would work by adding a link each of the current workers listed on the `invite_worker_to_company.html` page which would lead the user to the new route.

There, after the GET request, the user can check on the name and id of the worker it is about to remove. By click on a "remove" button the POST request is triggered.

petsch0:
I think that this is a good way to go. I assume the invite_worker_to_company.html is currently the only place where the workers from a company are shown right?

sloschert:
Yes

PR #1104, opened 2024-10-29
Add remove_worker method to CompanyResult

sloschert:
See #1092.

Plan: 8ec03a5b-3dbe-465d-a533-4b3bfe16121b


arbeitszeit_flask/database/repositories.py
```diff
@@ -731,6 +731,15 @@ def add_worker(self, member: UUID) -> int:
...
         return companies_changed
 
+    def remove_worker(self, member: UUID) -> int:
+        companies_changed = 0
+        member = models.Member.query.filter(models.Member.id == str(member)).first()
+        assert member
+        for company in self.query:
+            companies_changed += 1
+            company.workers.remove(member)
+        return companies_changed
```
petsch0, line 741: I know it is done the same way in add_worker but I still wonder if there is and/or will be in the future, a use case where self.query can contain more than one company? Isn't this action always triggered from exactly one company?

arbeitszeit_flask/database/repositories.py
```diff
@@ -731,6 +731,15 @@ def add_worker(self, member: UUID) -> int:
             company.workers.append(member)
         return companies_changed
```
Amittai-Aviram, line 732: Sorry to  hold things up, but do we expect ever to use this return value?  It looks as if we never do for add_worker — e.g., in arbeitszeit/use_cases/answer_company_work_invite.py, we have this:
`            self.database_gateway.get_companies().with_id(invite.company).add_worker(invite.member)`
It does not hurt, of course.

Amittai-Aviram (approved):
I have one question, but I am still approving.  The test looks adequate.

arbeitszeit_flask/database/repositories.py
sloschert: Yes, it seems to me that we do not use it currently. But as you said, it does not hurt..

petsch0 (approved):

sloschert:
1xCR

PR #1109, opened 2024-11-04
Fix flaky use case test in `test_query_plans.py`

sloschert:
Testcase  `test_zero_total_results_when_single_plan_filtered_by_id` failed when the randomly generated plan uuid contained "123". Now we query a unique string.


petsch0 (approved):
lgtm, thanks!

Issue #949, opened 2024-02-11
API: Change our existing API routes to except json instead of form data


sloschert:
When I go the browser at 127.0.0.1:5000/api/v1/doc and exectute a trial call for the consumptions.liquid_means_consumption method the request body looks like this:

```
plan_id=123&amount=2
```

I would instead expect the body to look like this:

```
{ "plan_id": "123", amount: 2 }
```

We should consider changing our existing API routes so that json instead of form data is accepted before implementing further methods.

_Originally posted by @seppeljordan in https://github.com/arbeitszeit/arbeitszeitapp/pull/946#pullrequestreview-1874344340_

sloschert:
has been fixed by the above mentioned PR.

PR #1110, opened 2024-11-04
api to accept json instead of form data

petsch0:
see #949 

note that we probably should create an issue to replace flask-restx RequestParser completely at some point:
https://flask-restx.readthedocs.io/en/latest/parsing.html


tests/api/integration/test_login_company_api.py
```diff
@@ -26,13 +26,15 @@ def test_post_returns_400_with_correct_data_in_query_string_but_missing_data_in_
         response = self.client.post(
             self.url, query_string=dict(email=email, password=password)
         )
-        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.status_code, 415)
+        # reponse shows expected content type
```
sloschert, line 30: What is the reason for this comment (here and in `test_login_member_api.py`)?

petsch0: I added it since I was not sure if it is common to all error codes that the response content type shows the expected content type, or if it is rather specific to 415 error code.

sloschert: I think this info is not only returned when there was an error, but also on success. We tested it on several occasions (search `self.assertEqual(response.mimetype, "application/json")`), and I guess it's not really necessary that you test it here. If you want to leave the test I'd suggest that you delete the comment (because the line of code is self-explaining) and/or move this check into a separate test function.

sloschert (approved):
One thing could be improved, I approve already. Thanks!

tests/api/integration/test_login_company_api.py
petsch0: you are right, I removed the comments

PR #1113, opened 2024-11-06
Allow QueryString objects to yield multiple values for one key

seppeljordan:
This commit changes the interface to `QueryString` objects. Before this change `QueryString` objects were only able to yield a single value for a given key. Now the method `QueryString.get` returns a list of all values supplied for a given key. The `get_last_value` method was added for convenient access to (possible) single value for cases where the app does not need to handle multiple for a single key.

This change was made because the old interface to the query string part of a URI was overly simplistic. It hid the fact that a query string can have multiple values for a single key from the developer. The new interface makes easier to remember the fact that a query string does not need to have only a single value for a key and makes it explicit that we choose a particular value from a possible set of values, which is currently always the last specified value in a query string.

### DIFF
--- a/arbeitszeit_web/request.py
+++ b/arbeitszeit_web/request.py
@@ -16,8 +16,20 @@ class Request(Protocol):
 
 
 class QueryString(Protocol):
-    def get(self, key: str) -> Optional[str]:
-        """Get the value for a particular query arguments."""
+    def get(self, key: str) -> list[str]:
+        """Get all values supplied by the client for a given key.
+
+        Example: Calling `r.get('a')` on a request with the query string
+            `a=1&a=2` will result in the list `['1', '2']`.
+        """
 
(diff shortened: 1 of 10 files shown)


arbeitszeit_web/request.py
```diff
@@ -16,8 +16,16 @@ def get_request_target(self) -> str: ...
 
 
 class QueryString(Protocol):
-    def get(self, key: str) -> Optional[str]:
+    def get(self, key: str) -> list[str]:
         """Get the value for a particular query arguments."""
```
petsch0, line 20: this could maybe also be updated with sth like
"Get the list of values for a query argument or [] if key does not exist." ?

petsch0 (approved):
good point, thanks! just one minor comment, and one question: do you already have a use case in the app in mind where sth like a=1&a=2 could be applied?

PR #1117, opened 2024-11-07
Refactor `EndCooperatoinController` to receive request as argument

seppeljordan:
Before this commit the `EndCooperationController` would receive request objects to process via its __init__ method. Although this worked it was suboptimal with regards to transparency and code reusability. Receiving request objects through the constructor limits the lifetime of the `EndCooperationController` to the lifetime of a request objects. The refactored implementation receives the request object as an argument to the `process_request_data` method.

The session object that is required by the controller to do its thing remains an argument to the `__init__` method and remains to be refactored in the same way as the request objects was in this change.


sloschert (approved):
Great change.

PR #1130, opened 2024-11-11
Refactor QueryPlansApiController

seppeljordan:
This commit changes how the QueryPlansApiController receives request objects to process. Before this commit the controller in question would receive request objects via a parameter of the `__init__` method. Now request objects are passed into the `create_request` method of the controller.


petsch0 (approved):
lgtm 👍  thanks @seppeljordan!

PR #1131, opened 2024-11-11
Refactor EndCooperationPresenter

seppeljordan:
This commit changes how the EnCooperationPresenter receives request objects that its supposed to generate responses for. Before this commit the presenter class would receive Request instances via a paramter of the `__init__` method of the presenter class. Now it receives request objects via a paramter of the `present` method.

### DIFF
--- a/arbeitszeit_web/www/presenters/end_cooperation_presenter.py
+++ b/arbeitszeit_web/www/presenters/end_cooperation_presenter.py
@@ -17,13 +17,14 @@ class EndCooperationPresenter:
         show_404: bool
         redirect_url: str
 
-    request: Request
     notifier: Notifier
     url_index: UrlIndex
     translator: Translator
     session: Session
 
-    def present(self, response: EndCooperationResponse) -> ViewModel:
+    def present(
+        self, response: EndCooperationResponse, *, web_request: Request
+    ) -> ViewModel:
         if response.is_rejected:
             self.notifier.display_warning(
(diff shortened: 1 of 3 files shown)


arbeitszeit_web/www/presenters/end_cooperation_presenter.py
```diff
@@ -17,13 +17,14 @@ class ViewModel:
...
 
-    request: Request
     notifier: Notifier
     url_index: UrlIndex
     translator: Translator
     session: Session
 
-    def present(self, response: EndCooperationResponse) -> ViewModel:
+    def present(
+        self, response: EndCooperationResponse, *, web_request: Request
```
petsch0, line 26: shouldn't we use named parameters everywhere by convention? why do you like to start forcing it here?

petsch0 (approved):
👍 just a minor question in this PR, and a general question wrt the intention of your latest "request refactorings": was it just to make it more explicit where the request object comes from, or to also untie the lifetimes of the objects from each other, or was it because of mocking?

petsch0:
I think you are definitely on the right path here :) I regularly see python test issues where there are so many dependencies and setup functions involved where some global objects are created in a specific way, then maybe mutated by a test case, and then not properly reset during several tear down methods, or not reset at all by the testcase itself.  

So, do I understand correctly, in #1132 a single test case calls FakeRequest(query_string=query_string) several times, but it will always be same object that is mutated right?

PR #1132, opened 2024-11-11
Refactor QueryCompaniesApiController tests

seppeljordan:
This commit groups together similar tests in the test suite for the QueryCompaniesApiController class.  The substance of the tests should preserved with the tst coverage being the exact same as before.


sloschert (approved):
:+1:

PR #1135, opened 2024-11-13
Create `RemoveWorkerFromCompanyPresenter`

sloschert:
On success, the presenter returns a redirect url to the "invite worker" route (where the company's workers are listed). Else, it returns an appropriate error code.

see #1092

### DIFF
--- a/arbeitszeit_web/www/presenters/remove_worker_from_company_presenter.py
+++ b/arbeitszeit_web/www/presenters/remove_worker_from_company_presenter.py
@@ -1,0 +1,47 @@
+from dataclasses import dataclass
+
+from arbeitszeit.use_cases.remove_worker_from_company import Response as UseCaseResponse
+from arbeitszeit_web.notification import Notifier
+from arbeitszeit_web.translator import Translator
+from arbeitszeit_web.url_index import UrlIndex
+from arbeitszeit_web.www.response import Redirect
+
+
+@dataclass
+class ErrorCode:
+    code: int
+
+
+ViewModel = Redirect | ErrorCode
+
+
+@dataclass
+class RemoveWorkerFromCompanyPresenter:
+    notifier: Notifier
+    translator: Translator
+    url_index: UrlIndex
+
+    def present(self, use_case_response: UseCaseResponse) -> ViewModel:
+        if not use_case_response.is_rejected:
+            self.notifier.display_info(
+                self.translator.gettext("Worker successfully removed from company.")
+            )
+            return Redirect(url=self.url_index.get_invite_worker_to_company_url())
+
+        assert use_case_response.rejection_reason is not None
+
+        match use_case_response.rejection_reason:
+            case UseCaseResponse.RejectionReason.company_not_found:
+                message = "Company not found."
+                code = 404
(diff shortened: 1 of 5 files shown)


arbeitszeit_web/www/presenters/remove_worker_from_company_presenter.py
```diff
@@ -0,0 +1,47 @@
...
+
+from arbeitszeit.use_cases.remove_worker_from_company import Response as UseCaseResponse
+from arbeitszeit_web.notification import Notifier
+from arbeitszeit_web.translator import Translator
+from arbeitszeit_web.url_index import UrlIndex
+from arbeitszeit_web.www.response import Redirect
+
+
+@dataclass
+class ErrorCode:
```
petsch0, line 11: do you really need a class for this single int? how about using a type alias instead?

arbeitszeit_web/www/presenters/remove_worker_from_company_presenter.py
```diff
@@ -0,0 +1,47 @@
...
+    code: int
+
+
+ViewModel = Redirect | ErrorCode
```
petsch0, line 15: just fyi I discussed this with @seppeljordan a little bit in person. To me it seemed strange that you do a type alias with an or here (a union as I learned), only to ensure type hinting is not complaining when an instance of 'ViewModel' is returned. Already the fact that it is even possible in terms of polymorphism (ViewModel can be of type Redirect or ErrorCode and these two classes do not have anything in common!?) confused me. 
But as I learned this is useful later for matching as for example in [register_private_consumption.py](https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/master/arbeitszeit_flask/views/register_private_consumption.py). Maybe we can talk about this again in a meeting. At least for me the explanation by @seppeljordan was very instructive, so thanks again :)

arbeitszeit_web/www/presenters/remove_worker_from_company_presenter.py
petsch0, line 11: do you really need a class for this single int? how about using a type alias instead?

petsch0 (approved):
thanks @sloschert! just a minor comment so I'll leave it up to you to decide if you want to change this PR again.

arbeitszeit_web/www/presenters/remove_worker_from_company_presenter.py
sloschert: Actually I prefer it to be a dataclass... I guess implementing it as a type alias would be like that:

```
ErrorCode = int

ViewModel = Redirect | ErrorCode
```

The small problem I have with this: When I call the presenter method and get the ViewModel, I can use `view_model.url` if ViewModel is a `Redirect` instance, but I cannot use `view_model.code` if it is a `ErrorCode` instance.  Don't really know if it is a convincing reason to reject your proposel ;-)

Btw, note that also the `Redirect` class is a dataclass with a single field:

```
@dataclass
class Redirect:
    url: str
```

sloschert: You are right, it is just as type safe. I prefered the dataclass because it is similar to the Redirect class. If you think i should change it let me know.

PR #1137, opened 2024-11-17
Small improvements to our test framework

sloschert:
**Improve login of users in our integration test framework** 

Before this change it was possible to specify member (or company) as well as password and email, when loging in users in the `ViewTestCase` of our test framework. This was confusing, because the user's credentials might not be compatible with the specified user id. The design has been improved.

**Let ViewTestCase.login_company return a UUID** 

Before this change the helper method `ViewTestCase.login_company`, used in many places of our integration tests, returned a `Company` record object, now it returns a UUID.

### DIFF
--- a/tests/flask_integration/flask.py
+++ b/tests/flask_integration/flask.py
@@ -7,7 +7,6 @@ from flask import Flask, current_app
 from flask_sqlalchemy import SQLAlchemy
 
 from arbeitszeit.injector import Module
-from arbeitszeit.records import Company
 from arbeitszeit_flask.database.repositories import DatabaseGatewayImpl
 from arbeitszeit_flask.token import FlaskTokenService
 from tests.data_generators import (
@@ -181,7 +180,7 @@ class ViewTestCase(FlaskTestCase):
         password: Optional[str] = None,
         email: Optional[str] = None,
         confirm_company: bool = True,
-    ) -> Company:
+    ) -> UUID:
         if password is None:
             password = "password123"
         if email is None:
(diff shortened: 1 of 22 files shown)


tests/flask_integration/flask.py
```diff
@@ -125,9 +123,9 @@ def login_member(
             password = "password123"
         if email is None:
             email = self.email_generator.get_random_email()
-        if member is None:
-            member = self.member_generator.create_member(
-                password=password, email=email, confirmed=False
+        if not self.database_gateway.get_members().with_email_address(email).first():
```
petsch0, line 126: just a suggestion to avoid having two db calls (here and l. 144): wouldn't it be possible to have a return var (sth like member _id) and then do sth like
`member = self.database_gateway.get_members().with_email_address(email).first()`
`if not member {`
`    member_id = self.member_generator.create_member`
`} else {`
`    member_id = member.id`
`}`
and then later just return member_id? just from looking at the code, did not test if that works though.

tests/flask_integration/flask.py
```diff
@@ -166,23 +169,25 @@ def login_accountant(self) -> UUID:
...
-    ) -> Company:
+    ) -> UUID:
         if password is None:
             password = "password123"
         if email is None:
             email = self.email_generator.get_random_email()
-        if company is None:
-            company = self.company_generator.create_company(
-                password=password, email=email, confirmed=False
+        if not self.database_gateway.get_companies().with_email_address(email).first():
```
petsch0, line 188: same as above

tests/flask_integration/flask.py
sloschert: You are right. The second db call was still from the times when we did not simply return a UUID, but a record object, that had to be recreated from the db before returning it.

petsch0 (approved):
lgtm, thanks @sloschert!

PR #1138, opened 2024-11-18
Create flask layer to remove worker from company

### LINKED ISSUE #1092: RFC Delete worker from company

sloschert:
_To be reviewed after PRs #1135 and #1136_

**Changes to RemoveWorkerFromCompanyController** 

    - display warning if invalid or no worker id
    - retrieve worker id from form, not query string
    - update controller tests
    - rename controller test module (add `test_` prefix)

**Create flask layer to remove worker from company** 

The now fully implemented workflow works as follows: Logged-in companies can click on a link on their `invite_worker_to_company.html` page in order to visit a new route, `remove_worker_from_company` (GET and POST). After the GET request, the company can select from a list of its workers to delete one. When clicking the submit button the POST request gets triggered that removes the worker.

fixes #1092


arbeitszeit_flask/templates/company/invite_worker_to_company.html
```diff
@@ -46,11 +46,15 @@ <h1 class="title">{{ gettext("Workers of the company") }}</h1>
...
             <tr>
               <td>{{ worker.id }}</td>
-              <td>{{ worker.name}}</td>
+              <td>{{ worker.name}}</td>                
             </tr>
             {% endfor %}
           </table>
         </div>
+        <h1 class="title">{{ gettext("Remove workers") }}</h1>
+        <div>{{ gettext("Visit this link to remove a worker:") }}
```
Amittai-Aviram, line 55: s/Visit this link/Click here/  (? — One does not actually visit a link, one visits a page by following a link.  I know, tiny and fussy point.)

arbeitszeit_flask/templates/company/remove_worker_from_company.html
```diff
@@ -0,0 +1,42 @@
...
+      <div>
+        <form method="post">
+          <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
+          <div class="field">
+            <label class="label">{{ gettext("Worker") }}</label>
+            <div class="control">
+              <div class="select">
+                <select name="worker">
+                  {% for worker in workers %}
+                  <option value="{{ worker.id }}">{{ worker.name }}</option>
```
Amittai-Aviram, line 25: Should the worker ID perhaps be displayed along with the worker's name?  What if the company has two people with the same name (Max Mustermann)?

arbeitszeit_flask/templates/company/remove_worker_from_company.html
```diff
@@ -0,0 +1,42 @@
...
+                  {% endfor %}
+                </select>
+              </div>
+            </div>
+          </div>
+          <div class="field">
+            <div class="control">
+              <button class="button is-danger" type="submit">{{ gettext("Remove worker") }}</button>
```
Amittai-Aviram, line 33: Does the user get a confirmation prompt?  If so, could the prompt contain the worker's name and ID number?

arbeitszeit_web/www/controllers/remove_worker_from_company_controller.py
```diff
@@ -2,18 +2,26 @@
...
+    translator: Translator
+
     def create_use_case_request(
         self, *, web_request: Request, session: Session
     ) -> UseCaseRequest | None:
         company = self._extract_company_from_session(session)
         worker = self._extract_worker_from_request(web_request)
         if worker is None:
+            self.notifier.display_warning(
+                self.translator.gettext("Invalid or no worker id in request.")
```
Amittai-Aviram, line 23: "Worker ID in request is invalid or missing."

arbeitszeit_web/www/presenters/remove_worker_from_company_presenter.py
```diff
@@ -0,0 +1,47 @@
...
+@dataclass
+class RemoveWorkerFromCompanyPresenter:
+    notifier: Notifier
+    translator: Translator
+    url_index: UrlIndex
+
+    def present(self, use_case_response: UseCaseResponse) -> ViewModel:
+        if not use_case_response.is_rejected:
+            self.notifier.display_info(
+                self.translator.gettext("Worker successfully removed from company.")
```
Amittai-Aviram, line 27: Let's see — the UseCaseResponse object here does not have identifying information about the worker, does it?  If it does, it would be good for the notifier to display it here as part of the string.

Amittai-Aviram (commented):
Looks great!  I am requesting a couple of tiny changes to some strings.  There are also a couple of small questions.

arbeitszeit_web/www/presenters/remove_worker_from_company_presenter.py
sloschert: No, it hasn't... Would complicate the response logic slightly. Let me know if you would like to have it implemented, than I'll do it.

arbeitszeit_flask/templates/company/remove_worker_from_company.html
sloschert: see presenter comment...

arbeitszeit_web/www/presenters/remove_worker_from_company_presenter.py
sloschert: I implemented the rest of your requests, thanks!

Amittai-Aviram: > No, it hasn't... Would complicate the response logic slightly. Let me know if you would like to have it implemented, than I'll do it.

No, that's OK — thank you!  Do you have time to write it up as a separate issue?  Somebody could go back later and fix it.  I think it would be much better if both the company removing the worker and the worker being removed got confirmations of that action, where the company would know exactly who was removed.  Since removal seems to me pretty serious — and permanent — the system should help the company to avoid mistakes.

Amittai-Aviram (approved):
LGTM.  Thanks for making those changes!  Please see my comment about writing an issue up regarding the company's confirmation message.  Thanks!

sloschert:
> LGTM. Thanks for making those changes! Please see my comment about writing an issue up regarding the company's confirmation message. Thanks!

I wrote an issue. Thanks for reviewing.

PR #1139, opened 2024-11-18
Remove usage of flask restx `RequestParser`

### LINKED ISSUE #1111: Remove flask-restx RequestParser from our code
We should probably replace deprecated flask-restx RequestParser at some point:
https://flask-restx.readthedocs.io/en/latest/parsing.html

Originally mentioned here:
https://github.com/ida-arbeitszeit/arbeitszeitapp/pull/1110#issue-2632525860

sloschert:
**Rename Api FormParameter to BodyParameter** 

This is a better name for the parameter in question. Its location is the body of a request and is to be distinguished from a Form parameter. See swagger v2 specs.

**Refactor Api docu decorator** 

This commit refactors the decorator `with_input_documentation`, which is responsible for generating the API documentation of the expected parameters. Now the decorator does not rely on the deprecated flask restx `RequestParser` anymore.

fixes #1111 

_FYI, a good tool to check the generated swagger.json document (http://127.0.0.1:5000/api/v1/swagger.json) for errors is https://editor.swagger.io/_

### DIFF
--- a/arbeitszeit_flask/api/input_documentation.py
+++ b/arbeitszeit_flask/api/input_documentation.py
@@ -31,48 +31,43 @@ class with_input_documentation:
     def __call__(self, func: Callable) -> Callable:
         """See https://swagger.io/specification/v2 for more information on the OpenAPI specification."""
         params: Dict[str, Any] = {}
-        for expected_input in self._expected_inputs:
-            if isinstance(expected_input, PathParameter):
-                self._add_path_parameter(params, expected_input)
-            elif isinstance(expected_input, QueryParameter):
-                self._add_query_parameter(params, expected_input)
-            else:
-                assert isinstance(expected_input, BodyParameter)
-                self._add_body_parameter(params, expected_input)
+        for input_ in self._expected_inputs:
+            match input_:
+                case PathParameter():
+                    params[input_.name] = {
+                        "description": input_.description,
+                        "type": "string",
+                        "required": True,
+                        "in": "path",
+                    }
+                case QueryParameter():
+                    params[input_.name] = {
+                        "description": input_.description,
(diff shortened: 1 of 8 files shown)


arbeitszeit_flask/api/input_documentation.py
```diff
@@ -1,54 +1,78 @@
...
     def __call__(self, func: Callable) -> Callable:
-        parser = RequestParser()
+        """See https://swagger.io/specification/v2 for more information on the OpenAPI specification."""
+        params: Dict[str, Any] = {}
         for expected_input in self._expected_inputs:
             if isinstance(expected_input, PathParameter):
-                # RequestParser does not handle path parameters
-                # https://github.com/noirbizarre/flask-restplus/issues/146
-                self._add_documentation_for_path_variable(func, expected_input)
+                self._add_path_parameter(params, expected_input)
```
petsch0, line 36: I am not sure if a separate function for each type is really required/helpful here (especially because the function is just an insert into a dict). I thought one advantage of those Unions would be that you could match them, so why not do this and then directly assign as done in the single funcs? Or, maybe have on "add" func that takes a "Parameter" type parameter and then does the type check?

sloschert: Consider the last commit I just pushed. I am not sure it it is better now, because we have now a single method that is quite long. But I have no strong opinions on this case, so if you prefer it that why, I am happy to merge it.
Let me also know, if I misunderstood your proposel.

petsch0: this is exactly what I meant. To me it is easier to read and extend, but maybe this is really personal preference. I think this is great work so I am going to approve the PR.

petsch0 (approved):
👍

Issue #825, opened 2023-09-22
Reverse (storno) registration of working hours


sloschert:
@seppeljordan @Amittai-Aviram @philipbroistedt 

Hi,

I think should agree on a general strategy how to deal with user mistakes like the following:

- Company registers wrong consumption (wrong plan id, amount or consumtion type)
- company registers wrong worked hours (wrong amount or worker)
- member registers wrong consumption (wrong plan id or amount)
- …

The easiest way to deal with those errors from a programmer perspective would be by trying to prevent them by issuing warnungs/confirmation dialogs (“Are you sure?…”). I think it would only need some (pretty generic) html and javascript.

But as described in the following article, this is not a satisfactory solution from a user perspective and “undo” mechanisms should be preferred: https://alistapart.com/article/neveruseawarning/

Given that i expect the implementation of proper “undo” mechanisms to be quite complex and consume a lot of time, i suggest the following as a general strategy:

- for the next milestone (“app is ready to use/feature complete”), we implement warnings/confirmation dialogs where needed
- After that milestone we add proper “undo” mechanisms where needed

Any thoughts? Especially: Can we consider the app feature complete without undo mechanisms for consumption registration?

petsch0:
If I recall correctly, outcome of our latest discussion was to start with kind of a minimal approach by implementing a proper undo mechanism as described above, but only for the use case
- company registers wrong worked hours (wrong amount or worker)
and maybe even split this up into multiple PRs?
From there we can discuss on how to continue for more complex use cases.

sloschert:
Yes, I think that's correct. 
Multiple PRs are always a good idea. E.g. first start with business logic, database methods or controller/presenters and only in the end write the front end code (views, routes, html) that triggers business logic, database actions, etc.

sloschert:
Going forward, I suggest the following use cases to be implemented.

1) use case “company can list its previously  registered working hours” 
2) use case “company can reverse a specific registration of working hours”
3) use case “server admin can restrict the time a company has to reverse previously registered working hours”

These use cases are high level descriptions, each step will probably require several changes and pull requests.

PR #1155, opened 2025-01-06
Undo register working hours

petsch0:
part of #825


tests/www/presenters/test_cancel_hours_worked_presenter.py
```diff
@@ -0,0 +1,68 @@
...
+
+class PresenterTests(BaseTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.presenter = self.injector.get(CancelHoursWorkedPresenter)
+
+    def test_that_user_gets_redirected_to_registered_hours_worked_via_referer_header(
+        self,
+    ) -> None:
+        expectedRedirectUrl = "/list_registered_hours_worked"
```
sloschert, line 21: The variable name does not follow python naming conventions.

arbeitszeit_web/www/presenters/cancel_hours_worked_presenter.py
```diff
@@ -0,0 +1,38 @@
...
+        if not use_case_response.delete_succeeded:
+            self.notifier.display_warning(
+                self.translator.gettext("Failed to delete registered working hours")
+            )
+        else:
+            self.notifier.display_info(
+                self.translator.gettext("Registered working hours successfully deleted")
+            )
+        return ViewModel(
+            redirect_url=request.get_header("Referer")
```
sloschert, line 36: Why do you think this line is necessary? Don't we always want to redirect to the get_registered_hours_worked_url in case of success?

tests/www/presenters/test_cancel_hours_worked_presenter.py
```diff
@@ -0,0 +1,68 @@
...
+    ) -> None:
+        expectedRedirectUrlIndex = "/get_registered_hours_worked_url"
+        request = FakeRequest()
+        request.set_header(key="Referer", value="")
+        response = self.get_response()
+        view_model = self.presenter.render_response(
+            use_case_response=response, request=request
+        )
+        self.assertIn(expectedRedirectUrlIndex, view_model.redirect_url)
+
```
sloschert, line 41: I think it would be better to not hardcode the url from the url index in this testcase but instead do something like this:

```
def test_that_user_gets_redirected_to_registered_hours_worked_via_url_index(
        self,
    ) -> None:
        request = FakeRequest()
        request.set_header(key="Referer", value="")
        response = self.get_response()
        view_model = self.presenter.render_response(
            use_case_response=response, request=request
        )
        assert view_model.redirect_url == self.url_index.get_registered_hours_worked_url()
```

tests/www/presenters/test_cancel_hours_worked_presenter.py
```diff
@@ -0,0 +1,68 @@
...
+from tests.request import FakeRequest
+from tests.www.base_test_case import BaseTestCase
+
+
+class PresenterTests(BaseTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.presenter = self.injector.get(CancelHoursWorkedPresenter)
+
+    def test_that_user_gets_redirected_to_registered_hours_worked_via_referer_header(
```
sloschert, line 18: The presenter will redirect to any url set as "Referer" in the header. IMO it would be less confusing to express this by changing the expected_redirect_url to something like `/any_url` instead of `/list_registered_hours_worked`.

tests/use_cases/repositories.py
```diff
@@ -1397,6 +1397,27 @@ def items() -> Iterable[Tuple[records.RegisteredHoursWorked, records.Member]]:
             database=self.database,
         )
 
+    def with_id(self, id_: UUID) -> Self:
+        return self._filter_elements(
+            lambda registered_hours_worked: registered_hours_worked.id == id_,
+        )
+
+    def delete(self) -> int:
```
sloschert, line 1405: I personally would find the following implementation a bit more straight foward, what do you think?

```
    def delete(self) -> int:
        registered_hours_to_delete = {entry.id for entry in self.items()}
        original_count = len(self.database.registered_hours_worked)

        self.database.registered_hours_worked = [
            entry
            for entry in self.database.registered_hours_worked
            if entry.id not in registered_hours_to_delete
        ]

        nr_of_entries_deleted = original_count - len(
            self.database.registered_hours_worked
        )
        return nr_of_entries_deleted
```

tests/flask_integration/database_gateway_impl/test_registered_hours_worked_result.py
```diff
@@ -184,6 +184,45 @@ def test_that_result_joined_with_worker_has_correct_worker_id(self) -> None:
...
+            .first()
+        )
+        self.assertEqual(records[0], result)
+
+    def test_that_no_record_is_returned_when_filter_by_random_id(self) -> None:
+        worker = self.member_generator.create_member()
+        company = self.company_generator.create_company(workers=[worker])
+        self.register_hours_worked(company=company, worker=worker, hours=Decimal(10))
+        self.assertIsNone(
+            self.database_gateway.get_registered_hours_worked().with_id(uuid4()).first()
```
Amittai-Aviram, line 206: But isn't it theoretically possible, just unlikely, that a random UUID could happen to match a valid UUID in the database?  If it is possible (even though unlikely), what would happen in that case?  Would that lead to an unreproducible test failure?

Amittai-Aviram (commented):
Looks great!  I have a couple of small questions.  Thank you!

tests/www/presenters/test_cancel_hours_worked_presenter.py
petsch0: done

arbeitszeit_web/www/presenters/cancel_hours_worked_presenter.py
petsch0: you are right, I removed the referer stuff

tests/www/presenters/test_cancel_hours_worked_presenter.py
petsch0: done

tests/www/presenters/test_cancel_hours_worked_presenter.py
petsch0: as the referer stuff was removed, I also removed this test

tests/use_cases/repositories.py
petsch0: done

petsch0:
> Great code, thanks! I left a few comments.
> 
> Moreover, it is optional, but it would be great if you could merge/rebase some or all of the commits in this PR for having a clearer git log? Or are you planning to do a squash merge?

yes, I would squash merge this if it is okay. Do you mean "clearer git log" because of the merge commits in-between?

> 
> Also, neither super important, but a thing I have noticed: We informally agreed in the past to use the "pytest" way of asserting statements in tests. For example, we try to use `assert variable == "expected"` instead of `self.assertEqual(variable, "expected")`. But if you prefer the unittest way, it's fine for me.

good point, done

arbeitszeit/use_cases/cancel_hours_worked.py
```diff
@@ -0,0 +1,35 @@
...
+class CancelHoursWorkedUseCase:
+    database: DatabaseGateway
+    datetime_service: DatetimeService
+
+    def cancel_hours_worked(self, request: Request) -> Response:
+        requester = self.database.get_companies().with_id(request.requester).first()
+        query_result = self.database.get_registered_hours_worked().with_id(
+            request.registration_id
+        )
+        row_to_be_deleted = query_result.first()
```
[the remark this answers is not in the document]
petsch0: by signature, with_id returns RegisteredHoursWorkedResult, so I guess no, it will never be None. However, first() CAN return None and is therefore checked afterwards. It is done the same way in other use cases.

tests/flask_integration/database_gateway_impl/test_registered_hours_worked_result.py
petsch0: I do not think this is an issue, especially not in this test case. we are using uuid4() all over the app not just in tests. It is guaranteed to not generate duplicates afaik. If that was not the case, the whole db table setup would be questionable right?

sloschert (approved):

PR #1162, opened 2025-01-14
Allow admins to configure alembic.ini config file path

seppeljordan:
This commit allows admins/users of the application to provide a custom alembic config file for their installation of the arbeitszeitapp.

### DIFF
--- a/arbeitszeit_flask/__init__.py
+++ b/arbeitszeit_flask/__init__.py
@@ -1,5 +1,4 @@
 import os
-from pathlib import Path
 from typing import Any
 
 from alembic import command as alembic_command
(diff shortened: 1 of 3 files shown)


sloschert (commented):
I wonder if this is really something that we should completely leave to the admins. Wouldn't it be quite some work for an admin to dig into the alembic docs to find out how to make the config working? Why can't we just ship our alembic.ini file? Or shouldn't we provide at least a default alembic.ini, that can then be modified?

sloschert (commented):
Or am I misunderastanding your approach and you are planning to ship an alembic.ini file?

sloschert (approved):
As discussed, I approve.

sloschert:
@seppeljordan see https://github.com/ida-arbeitszeit/arbeitszeitapp/pull/1163

PR #1166, opened 2025-01-30
Fix bug in auto migration

sloschert:
Currently, on startup, all tables of the db get created via `Base.metadata.create_all` and afterwards, if AUTO_MIGRATE is set true, alembic tries to upgrade the db. This will fail, because the tables exist already. By reversing the order, this bug get fixed, because `Base.metadata.create_all` does not try to recreate a table if it exists already.


sloschert:
I detected a problem, will close it for now.

PR #1170, opened 2025-02-01
Fix bug in auto migration

sloschert:
**Add test for startup w/ auto_migrate = True**

**Fix bug in auto migration** 

Before this commit, the AUTO_MIGRATE config option did not work as expected for the test or the production db, because our migration tool alembic would migrate always the dev db. Now the correct db engine is passed to the migration script in `arbeitszeit_flask.migrations.env.py`, to make sure that the correct db is auto migrated.

Moreover, when auto migrating, alembic now "stamps" databases that were not initialized with alembic, to enable the migration on that db.

### DIFF
--- a/arbeitszeit_flask/migrations/env.py
+++ b/arbeitszeit_flask/migrations/env.py
@@ -4,9 +4,6 @@ from arbeitszeit_flask import load_configuration
 
 from arbeitszeit_flask.database.db import Database
 from flask import Flask
-from collections.abc import Iterable
-from alembic.environment import MigrationContext
-from alembic.operations import MigrationScript
 
 from alembic import context
 
@@ -14,15 +11,11 @@ from alembic import context
 # access to the values within the .ini file in use.
 config = context.config
 
-# Interpret the config file for Python logging.
 # This line sets up loggers basically.
 if config.config_file_name is not None:
     fileConfig(config.config_file_name)
 
-# add your model's MetaData object here
-# for 'autogenerate' support
-# from myapp import mymodel
-# target_metadata = mymodel.Base.metadata
+# 'autogenerate' support
 from arbeitszeit_flask.database.models import Base
 target_metadata = Base.metadata
 
@@ -33,28 +26,7 @@ target_metadata = Base.metadata
 
 
 def run_migrations_offline() -> None:
-    """Run migrations in 'offline' mode.
-
-    This configures the context with just a URL
-    and not an Engine, though an Engine is acceptable
-    here as well.  By skipping the Engine creation
-    we don't even need a DBAPI to be available.
-
-    Calls to context.execute() here emit the given string to the
-    script output.
-
-    """
-    url = config.get_main_option("sqlalchemy.url")
-    context.configure(
-        url=url,
-        target_metadata=target_metadata,
-        literal_binds=True,
-        dialect_opts={"paramstyle": "named"},
-    )
-
-    with context.begin_transaction():
-        context.run_migrations()
-
+    raise NotImplementedError()
 
 def get_db_uri() -> str:
     tmp_flask_app = Flask(__name__)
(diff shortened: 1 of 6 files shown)


arbeitszeit_flask/migrations/env.py
```diff
@@ -33,68 +26,33 @@
...
-        url=url,
-        target_metadata=target_metadata,
-        literal_binds=True,
-        dialect_opts={"paramstyle": "named"},
-    )
-
-    with context.begin_transaction():
-        context.run_migrations()
-
+    raise NotImplementedError()
```
petsch0, line 29: why is this removed?

arbeitszeit_flask/migrations/env.py
```diff
@@ -33,68 +26,33 @@
...
-    """Run migrations in 'online' mode.
-
-    In this scenario we need to create an Engine
-    and associate a connection with the context.
-
-    """
-    # this callback is used to prevent an auto-migration from being generated
-    # when there are no changes to the schema
-    # reference: http://alembic.zzzcomputing.com/en/latest/cookbook.html
-    def process_revision_directives(
```
petsch0, line 76: from the description this sounds quite helpful. I checked the docs and it seems it still exists: https://alembic.sqlalchemy.org/en/latest/cookbook.html#don-t-generate-empty-migrations-with-autogenerate
shouldn't we continue to use it?

tests/flask_integration/flask.py
```diff
@@ -72,14 +72,16 @@ def setUp(self) -> None:
...
+        # At this point, the database has already been configured
+        # and its tables have been created via the `create_app` function in
+        # `arbeitszeit_flask/__init__.py`
+        self.db = Database()
 
     def tearDown(self) -> None:
         self._lazy_property_cache = dict()
         self.db.session.remove()
         self.app_context.pop()
+        Base.metadata.drop_all(self.db.engine)
```
petsch0, line 84: I guess that is to clean the db after a test run? how did this work before?

tests/flask_integration/test_db_migrations.py
```diff
@@ -0,0 +1,39 @@
...
+        return modules
+
+
+class AutoMigrationTests(AutoMigrationsTestCase):
+    @property
+    def auto_migrate_setting(self) -> bool:
+        return True
+
+    def test_that_app_starts_successfully_with_auto_migrate(self) -> None:
+        assert self.app
```
petsch0, line 39: I do not really understand how this works. maybe we can have briefly talk about it in the weekly?

petsch0 (commented):
just some questions since I do not completely understand what is going on in this PR. It is still okay for me though if you just go on and merge this if you think it is required.

arbeitszeit_flask/migrations/env.py
sloschert: Because we are not using this function, and I wanted to reduce unnecessary noise/complexity in that module. We are not calling this function, neither when we are upgrading/downgrading the dev databases manually via the `alembic` command nor when we update the test or prod databases via the AUTO_MIGRATE option. I think this gets clearer by deleting the function body. If we should need that functionality in the future we can add it.

arbeitszeit_flask/migrations/env.py
sloschert: I think it adds too much complexity to the code for the small advantage it brings. One can simply use `alembic check` in order to see if there are pending changes. This will not generate a migration file. I personally never called `alembic revision --autogenerate` without knowing that there are changes pending (and if I did I could simply delete the generated migration file).

tests/flask_integration/flask.py
sloschert: Yes, it deletes all the tables. Note that it does not clean types and does not drop the database, but it's clean enough for our purposes, I think. 
If I am not mistaken it worked before, because in the setup method of every test we ran these two lines in order to have a clean database:

```
Base.metadata.drop_all(self.db.engine)
Base.metadata.create_all(self.db.engine)
```

But i thought it's cleaner to erase the tables in the teardown method. 
And we do not have to call explicitly `Base.metadata.create_all(self.db.engine)` in the setup method, because the database gets initialized in the call to `self.app = self.injector.get(Flask)`.

tests/flask_integration/test_db_migrations.py
sloschert: Yes, sure. The pattern ensures that the flask app gets loaded with the correct, manipulated flask configuration needed for the current test. (That happens [here](https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/65c61f4df3da4c4fa31e737a4e82cea582685506/arbeitszeit_flask/__init__.py#L51))

You find exactly the same pattern in the testing modules [`tests/flask_integration/test_language_repository.py` ](https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/65c61f4df3da4c4fa31e737a4e82cea582685506/tests/flask_integration/test_language_repository.py#L10) and [`tests/flask_integration/test_request_email_change_view.py`](https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/65c61f4df3da4c4fa31e737a4e82cea582685506/tests/flask_integration/test_request_email_change_view.py#L88)

tests/flask_integration/flask.py
petsch0: I've been part in similar discussions a couple of times. Some say the "test framework" must guarantee a clean state when test case starts, others argue that it is your (the test cases') responsibility to ensure a clean state before running any test case. imo both approaches are fine :)

petsch0 (approved):
lgtm, nice work!

PR #1176, opened 2025-02-08
Delete unnecessary 'auth.zurueck' route.

### LINKED ISSUE #1161: Rename route "auth.zurueck"
The route "zurueck" in the "auth" blueprint should be renamed. The name should be in English and probably also made clearer, that the user gets logged out and redirected to the "start" page.

Maybe the whole route is unnecessary and can get deleted after some refactoring.

https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/a6fa3b4731ff0dad697098f1fbb672f5a8f8f5b8/arbeitszeit_flask/auth/routes.py#L265

sloschert:
Fixes #1161

### DIFF
--- a/arbeitszeit_flask/auth/routes.py
+++ b/arbeitszeit_flask/auth/routes.py
@@ -67,7 +67,6 @@ def set_language(language: str):
     return redirect(redirection_url)
 
 
-# Member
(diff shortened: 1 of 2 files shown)


petsch0 (approved):
lgtm!

PR #1177, opened 2025-02-09
Don't show purpose of wage transaction

### LINKED ISSUE #1165: Worker Transaction purpose shown in German ("Lohn") in all languages
The worker transaction overview in route `/member/my_account` shows the German word "Lohn" in all transactions of "wages" (work certificates). 

This word gets set in[ the business logic (use case) layer](https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/f82bcfc8b8ca15005f0bdb94bc4b71ff926cfe05/arbeitszeit/use_cases/register_hours_worked.py#L72) and should be translated in the adapter layer (presenter).

Image

sloschert:
The purpose was just the string "Lohn". Fixes #1165

### DIFF
--- a/arbeitszeit_flask/templates/macros/transactions.html
+++ b/arbeitszeit_flask/templates/macros/transactions.html
@@ -94,8 +94,10 @@
                     {{ type }}
                 </strong>
                 <br>
+                {% if purpose %}
                 <small>{{ purpose }}</small>
                 <br>
+                {% endif %}
                 <small>{{ user_name }}</small>
             </p>
         </div>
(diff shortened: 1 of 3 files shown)


petsch0:
What is the reason to completely remove the purpose field from wage transaction instead of simply translating it like other html strings in the app? I actually think it might become quite useful to further differentiate same transaction types, similar to  the "purpose of use" (Verwendungszweck) field for bank transactions.

sloschert:
> What is the reason to completely remove the purpose field from wage transaction instead of simply translating it like other html strings in the app? I actually think it might become quite useful to further differentiate same transaction types, similar to the "purpose of use" (Verwendungszweck) field for bank transactions.

Currently it does not provide any extra info, it's a duplication. And in general I think we should find better ways to identify transfer types than via the free text field "purpose" (e.g. via joins with other db tables that have a foreign key on transaction). IMO it was a bad decision to introduce this field in the first place and we should get rid of it.

Amittai-Aviram (approved):
I'm confused about one thing.  Right now, are we not actually using this field at all in the DB?  If we are, then don't we need to migrate the DB so as to remove that column?  If we are not, then — then it's really weird!  Anyway, otherwise, LGTM!

Issue #1175, opened 2025-02-03
RFC - Design of accounts and transactions


sloschert:
In IDA discussions we came to the conclusion that in a labor time economy no hours should simply „disappear“ or „appear out of nowhere“. I.e., every transfer of hours must be tracked in _accounts_, because they enable accountability: An account has a purpose and responsibles.

IMO, a _transaction_ should have the following attributes:
- sending account
- receiving account
- value (in hours; there must be no difference between sent and received value)

In the attached flowchart I try to list all the relevant accounts and transactions that exist (or should exist) in our app. Two accounts are still missing and several transactions have to be added or changed:

**Missing accounts:**

- Public sector fund (PSF) account:
It should be an account of social accounting, in addition to the already existing general account of social accounting.

- Cooperation account:
Every cooperation should have such an account. Its purpose is to track the differences between planned prd and coop price for each consumption.  

**Transactions missing/to be changed:**

- In "approval of productive plan": Instead of social accounting sending a negative amount to Company PRD account, Company PRD account should send planned PRD to social accounting. 

- In "approval of public plan": PSF account should send planned PRD to social accounting general account.

- In "private consumption": Currently the difference between coop price and planned prd is nowhere explicitly tracked. There must be a transaction from company prd account to cooperation account OR vice versa, depending if the difference is positive or negative.

- In "productive consumption of fixed means of production": Currently the difference between coop price and planned prd is nowhere explicitly tracked. There must be a transaction from selling company prd account to cooperation account OR vice versa, depending if the difference is positive or negative.

- In "productive consumption of liquid means of production": Same as above.

- In "registration of hours worked": Currently the difference between the worked hours and the part deducted for the public sector, determined by the FIC, is not tracked in terms of transactions. There should be two transactions from the Company A account: One to the Member account (value: Worked hours * FIC) and one to the PSF account (value: Worked hours * (1 - FIC)). 

I am interested in your opinion if that design makes sense for you. @petsch0 @Amittai-Aviram 

Please have a look at this flowchart:

[accounts and transactions.pdf](https://github.com/user-attachments/files/18648272/accounts.and.transactions.pdf)

[accounts and transactions.odg](https://github.com/user-attachments/files/18648278/accounts.and.transactions.odg)

petsch0:
Thanks a lot @sloschert! I did not yet think this through, just some comments/questions from the top of my head:
[...]

PR #1178, opened 2025-02-10
Add dev docs section on 'accounts and transactions'

sloschert:
Note that adding details on transactions and our implemention is due after discussion of #1175

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -3,10 +3,24 @@ Concepts
 
 This document explains key concepts of the Arbeitszeitapp.
 
-Accounts
---------
 
-tbd.
+Accounts and Transactions  
+-------------------------
+
+The core principle of labor time accounting is to prevent the appropriation of others' work. Therefore, all 
+transactions of working hours must be recorded, and there must be clear rules regarding which transactions 
+are permitted. As proposed by the "Group of International Communists" (GIC), 
+this goal is achieved through bookkeeping of transactions between accounts.  
+
+According to the GIC, the following accounts are essential in labor time accounting:  
+
+- Each company must have one account for fixed means of production (P), one for liquid means of production 
+or raw materials (R), one for labor (A), and one for the total product (PRD).  
+- Each member must have an account for labor certificates.  
+
+Additionally, the following accounts are necessary to ensure consistent bookkeeping:  
+- Each cooperative must have a balancing account. 
+- Social accounting must have a Public Sector Fund (PSF).
 
 
 Plans


docs/concepts.rst
```diff
@@ -3,10 +3,24 @@ Concepts
...
+-------------------------
+
+The core principle of labor time accounting is to prevent the appropriation of others' work. Therefore, all 
+transactions of working hours must be recorded, and there must be clear rules regarding which transactions 
+are permitted. As proposed by the "Group of International Communists" (GIC), 
+this goal is achieved through bookkeeping of transactions between accounts.  
+
+According to the GIC, the following accounts are essential in labor time accounting:  
+
+- Each company must have one account for fixed means of production (P), one for liquid means of production
```
Amittai-Aviram, line 17: Style: put these four accounts as subordinate bullet points —
- Each company must have four accounts:
  - Fixed means of production (P)
  - Liquid means of production, or raw materials (R)
  - Labor (A, "Arbeit")
  - The total product (PRD), debited at the start of each plan cycle by the expected total value of the product.
- Each member must have ...

docs/concepts.rst
```diff
@@ -3,10 +3,24 @@ Concepts
...
+this goal is achieved through bookkeeping of transactions between accounts.  
+
+According to the GIC, the following accounts are essential in labor time accounting:  
+
+- Each company must have one account for fixed means of production (P), one for liquid means of production 
+or raw materials (R), one for labor (A), and one for the total product (PRD).  
+- Each member must have an account for labor certificates.  
+
+Additionally, the following accounts are necessary to ensure consistent bookkeeping:  
+- Each cooperative must have a balancing account.
```
Amittai-Aviram, line 22: What is a "balancing account"?  Can we be more precise?

sloschert: I'd like to delete the cooperative account from the docs until we proceeded with our discussion on it.

docs/concepts.rst
sloschert: Good idea to use subordinate bullet points. 

The description of the PRD account as proposed by you does not reflect that it is also used to track the consumptions of the product. I propose to either adapt it further or to leave it very minimal for now and expand it after proceeding with out discussion on accounts and transactions. I think after agreeing on an account and transaction model it will be much easier writing the detailed docs.

Amittai-Aviram: I am in favor of adapting it:
The total product (PRD), debited at the start of each plan cycle by the sum of the P, R, and A values, and then increased back toward zero, item by item, as the finished product gets consumed.

sloschert: Fine, changed it!

Amittai-Aviram (approved):
LGTM.  Yay!  Thank you for doing this!

Issue #1035, opened 2024-07-02
Vanilla Frontend Transformation Process 🍦🍦🍦


4lm:
In the last dev online meeting (Jun 25), we decided to give [my proposal in #989](https://github.com/ida-arbeitszeit/arbeitszeitapp/issues/989#issuecomment-2182921840) a go.

The goal of the in #989 described [vanilla](https://en.wikipedia.org/wiki/Vanilla_software) frontend transformation process (VFTP) is to remove third-party ballast like Bulma CSS and FontAwesome and instead use what the web platform has to over natively. Following, a list of our main goals:

- Remove third-party frontend dependencies (FontAwesome, Bulma)
- Use Semantic HTML tags & standardized HTML elements (W3C)
- [Accessibility](https://developer.mozilla.org/en-US/docs/Web/Accessibility) (A11Y) shall become a first-class citizen
- The app shall follow [progressive enhancement](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement) principles
- The layout must be customizable by the user via a custom CSS file
- The frontend design shall be more functional, useful, and less obtrusive

As requested in this meeting, I will outline a short action plan with checkboxes in the following section. Each checkbox will represent a PR. This issue will accompany the VFTP and, therefore, contain more and more checkbox-type entries over time. This more checkboxes over time approach aims to start fast and become more specific over time. Therefore, the following section is constantly WIP until the VFTP is finished:

- [ ] Remove FontAwesome (has a 623 KB JS dependency)
  - [ ] Replace all FontAwesome icons with their native SVG counterpart (with zero percent JS)
  - [ ] Write dev docs on how to implement SVG icons easily
  - [ ] Add a non-functional test that secures frontend data size
- [ ] Upfront, adjust frontend styling issues  that have nothing to do with Bulma (Bulma stays, for now)
  - [ ] Switch to Bulma min version (same feature set > 650 KB less CSS data to transfer)
[...]

Issue #937, opened 2024-02-02
Signup Form: Allow users to view the entered password in password field


seppeljordan:
Currently users cannot view the password they entered into the password field in the signup form. Other applications tend to offer a toggle switch that reveals the text entered into a password input field. We should have something like that in the signup form.

sloschert:
Accomplished by PR #1181

PR #1181, opened 2025-02-16
ui toggle to view entered pw

petsch0:
issue #937. i chose the usual "eye"/"eye-slashed" icon for that purpose. the javascript code is quite redundant, let me know if I should refactor it a bit. I thought it may not be worth it since it works and we still may want to continue with #1035.


arbeitszeit_flask/static/main.js
```diff
@@ -47,3 +47,33 @@ document.addEventListener('DOMContentLoaded', () => {
     });
 
   });
+
+// toggle password visibility
+document.addEventListener('DOMContentLoaded', () => {
+  $eye = document.getElementById("eye");
```
sloschert, line 53: probably teh variables here should be declared with const or let instead of implicitly creating a global variable.

arbeitszeit_flask/static/main.js
```diff
@@ -47,3 +47,33 @@ document.addEventListener('DOMContentLoaded', () => {
...
+  $eyeSlashed = document.getElementById("eye-slashed");
+  $pwInput = document.getElementById("password");
+  $eyeRepeat = document.getElementById("eye-repeat");
+  $eyeSlashedRepeat = document.getElementById("eye-slashed-repeat");
+  $pwInputRepeated = document.getElementById("repeat_password");
+    $eye.addEventListener('click', () => {
+      $eye.classList.add("is-hidden")
```
sloschert, line 60: Missing semicolon at the end of this line and below.

sloschert (approved):
Added two small comments. You can merge the PR directly after applying the changes, if you agree with them.

arbeitszeit_flask/static/main.js
petsch0: done, i changed the whole execution logic. please have another look.

arbeitszeit_flask/static/main.js
petsch0: afaik semicolons are optional in javascript, but I you insist I will add them :)

Amittai-Aviram (approved):
LGTM!  I guess we generally do not have any (automated) tests for GUI stuff of this sort, correct?

petsch0:
> LGTM! I guess we generally do not have any (automated) tests for GUI stuff of this sort, correct?

Yes, that is right, afaik there are no such tests yet. I guess that would have been another subtask of #1035

PR #1192, opened 2025-03-15
add worker id and name and email to worker to removal notification

### LINKED ISSUE #1145: Include worker id and name in company notification on successful removal of worker
When a company successfully removes a worker, it gets a notification. But this notification does not contain the name and id of that worker. This should be implemented.

---

I think it would be much better if both the company removing the worker and the worker being removed got confirmations of that action, where the company would know exactly who was removed.  Since removal seems to me pretty serious — and permanent — the system should help the company to avoid mistakes.

_Originally posted by @Amittai-Aviram in https://github.com/ida-arbeitszeit/arbeitszeitapp/pull/1138#discussion_r1867872907_

petsch0:
fixes #1145 

note that the current implementation has some drawbacks that should be addressed at some point. First, in arbeitszeit/use_cases/remove_worker_from_company.py::RemoveWorkerFromCompanyUseCase::_notify_worker_and_company, self.email_sender.send_email is called, but in the notify-implementation of arbeitszeit_web/email/notify_about_worker_removal_from_company_presenter.py::NotifyAboutWorkerRemovalPresenter actually two emails are sent. Separating this is possible, but it makes the second drawback even worse as the email sending in this use case should be tightly coupled imo. This second drawback comes down to the fact that afaics, email sending in the app is not checked for errors, i.e. it is possible that the first notification mail for the worker is is sent successfully, while the second one for the company fails, and this will go unnoticed (see [send_message](https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/f9991ddb05d40fd2aa7cef2396ecf929ab3cc111/arbeitszeit_flask/mail_service/smtp_mail_service.py#L17)). But even if email sent errors were checked, the question arises how to handle such a case, e.g. rollback the worker removal completely, just retry email sending with backoff, or something else?

Let me know what you think regarding this point, and if I should create an issue for it.


sloschert:
Thanks for working on this! When I wrote that issue I actually was not proposing to create a new email for companies, but to enhance the existing notification in the _presenter_ (which results in a flask flash message in our flask layer): 

https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/f9991ddb05d40fd2aa7cef2396ecf929ab3cc111/arbeitszeit_web/www/presenters/remove_worker_from_company_presenter.py#L26

Sorry for not being more clear on that. I tend to think that creating one company email for every worker removed is slightly overkill and that a enhanced info in the presenter notification should be good enough. 

Reasons why I think it would be overkill:
- It might be necessary e.g. for administative reasons to remove many of workers at the same time, which would result in as many emails in the company inbox
- The company user should basically know what they are doing. Name and id of the worker in a flash message should just allow a quick and basic check that they removed the right worker 
- The company can always check the list of their workers to see they made no mistake

What do you think?

P.S.: The problem of us not checking the success of email sendings is something worth dicussing (maybe logging?) but not prio 1 IMO, because the email config is under control of the respective app admin.

petsch0:
I agree with your P.S. point regarding successful email sending. 

Regarding #1145, I think we should discuss again. Because it includes also @Amittai-Aviram's comment, and a notification for the worker cannot be implemented with those flash messages right? 
Besides that, I do not think that those flash messages are very useful at all, and that they should be part of core functional implementations. I consider them "UI-sugar" (especially the green ones) and personally I tend to ignore them. I rather agree with @Amittai-Aviram's comment in #1145 on this point, i.e. that worker removal is a pretty serious action with possible real-life consequences, and that therefore, all involved parties should get a persisted confirmation (like an email) for their records. I also do not see an issue with sending out a 100 (or 200) emails, if 100 workers were removed at the same time (which is not possible with the current UI interface btw). I imagine such bulk operation would be done via API, and in this case email confirmations would become even more crucial as there may not be any other visual confirmation.

sloschert:
"notification for the worker cannot be implemented with those flash messages right?"

I did not question the mails to workers, i was wondering if we need a mail to the company.

petsch0:
ok, now I get it. I do not really see the benefit of adding data to a disappearing info box though, but I am okay with closing this PR then, and adding those fields to the info box, if you think this is sufficient.

sloschert:
The fact that these info boxes are disappearing was a one-person design decision a long time ago. Would it make things different for you if these info boxes would NOT disappear? Because we can certainly do this (but would need to fix some frontend details, because the advantage of the current implementation is that these disappearing info boxes can appear always at the top of the screen)

petsch0:
well, in general my argumentation from above still stands, i.e. I think that in that specific use case it is justified to send out mails to all involved parties. I would not want to change the behaviour of the info boxes in general just because of this single use case. I would rather question #1145 and simply close it. maybe it is worth discussing this in a weekly meeting.

arbeitszeit/use_cases/remove_worker_from_company.py
```diff
@@ -46,20 +46,31 @@ def remove_worker_from_company(self, request: Request) -> Response:
...
+        company_record = (
+            self.database_gateway.get_companies()
+            .with_id(workplace.id)
+            .joined_with_email_address()
+            .first()
+        )
+        if worker_record and company_record is not None:
+            _, worker_email = worker_record
+            _, company_email = company_record
             self.email_sender.send_email(
```
sloschert, line 68: My preference is to separate the two mail sending procedures, I think it would be more intuitive and enables more detailed use case testing. Have you seen the use case tests for email sending in `tests/use_cases/test_remove_worker_from_company.py`? In any case, these should be adapted to reflect the changes in the email(s) sent.

arbeitszeit/email_notifications.py
```diff
@@ -40,6 +40,9 @@ class WorkerInvitation:
 @dataclass
 class WorkerRemovalNotification:
     worker_email: str
+    worker_name: str
+    worker_id: str
```
sloschert, line 44: We are here still in the business logic and should use `UUID`s and not strings, where possible. No need to convert the worker_id to string.

arbeitszeit/use_cases/remove_worker_from_company.py
```diff
@@ -46,20 +46,31 @@ def remove_worker_from_company(self, request: Request) -> Response:
...
             return Response(Response.RejectionReason.not_workplace_of_worker)
         workplace_record.remove_worker(worker.id)
-        self._notify_worker(worker, workplace)
+        self._notify_worker_and_company(worker, workplace)
         return Response(None)
 
-    def _notify_worker(self, worker: Member, workplace: Company) -> None:
-        record = (
+    def _notify_worker_and_company(self, worker: Member, workplace: Company) -> None:
+        worker_record = (
```
sloschert, line 53: You might consider joining member and workplace in line 41 and 44 already with the respective email addresses, which would save us two calls to the database and the check if worker and company exist.

sloschert (changes requested):
Ok, for me it's totally fine if we send emails also to the company. I made some comments, please see my review.

Let's discuss the broader topic of emails vs. "presenter notifications" anyway in one or two meetings.

arbeitszeit/use_cases/remove_worker_from_company.py
Amittai-Aviram: Well — but what would be wrong with sending just a single e-mail message to the two recipients, the worker and the company?  That would simplify the application logic and leave the mechanics of sending to two recipients to the e-mail server,  (I do not mean to complicate your work, of course!)

petsch0:
taking this up again, thanks for your latest comments @sloschert @Amittai-Aviram! I liked your suggestion @Amittai-Aviram to simply send the same mail to both worker and company! No idea why I did not come up with this in the first place :smile:

sloschert:
> btw @sloschert after merging master into this branch, and running `alembic upgrade head && flask run` db migration failed with
> 
> > psycopg2.errors.DuplicateTable: FEHLER:  Relation »transfer« existiert bereits
> 
> so I had to manually delete the table. Also `test_db_migrations.py` is continuously failing locally.

after the changes to the test db setup and tear down logic in one of the last PRs, it might be necessary to drop the test db completely and recreated it. On linux `dropdb <test_db_name> && createdb <test_db_name>`

sloschert (approved):
Great work, thanks!

PR #1194, opened 2025-03-16
Validate accountant email matches the invited one in wt form

### LINKED ISSUE #1171: Poor error handling when accountant registers with wrong email address
**Summary**

When an invited accountant tries to register with a wrong email, the registration fails without notification.

**How to reproduce**

Invite an accountant with terminal command `flask invite-accountant example@mail.de`. Click the link in the invitation email. Try to register the accountant providing a different email address, e.g. `another_example@mail.de`

**Technical details**

The email comparison happens in the [register accountant controller](https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/65c61f4df3da4c4fa31e737a4e82cea582685506/arbeitszeit_web/www/controllers/register_accountant_controller.py#L21).

**Desired outcome**

The user should receive a notification stating that the provided email by the user does not match the email provided by the inviting organization.

petsch0:
fixes #1171 

I chose to add the required email check as part of the form validation instead of doing it in the controller, since I think that semantically it belongs there, and I also did not want to continue to "misuse" those flash messages for form validation. Downside is potentially that if one day it is decided to get rid of WT forms, then this check would have to be re-implemented.

Also note that the registration will continue to fail silently (as described in #1171) in case an invalid token is used as part of the url. So there is a difference in opening the correct link and then providing a different email in the form (results in validation error), versus directly opening a random url like http://127.0.0.1:5000/accountant/signup/test and then submitting this form (fails silently/no op). Let me know if you think this should be handled differently.


arbeitszeit_flask/forms.py
```diff
@@ -174,6 +176,16 @@ class RegisterAccountantForm(Form):
         ],
     )
 
+    def validate_email(form, field) -> None:
```
sloschert, line 179: you could add type annotations: `form: Self, field: StringField`

arbeitszeit_flask/forms.py
```diff
@@ -174,6 +176,16 @@ class RegisterAccountantForm(Form):
...
+        if field.data != form.extracted_token:
```
sloschert, line 180: what about making the email comparison case-insensitive and remove whitespaces?

```
input_email = field.data.lower().strip() if field.data else ""
token_email = form.extracted_token.lower().strip() if form.extracted_token else ""
    
if input_email != token_email:
    ...
```

sloschert (changes requested):
Thanks! Looks good, I only have two suggestions. 

So registration fails silently when the token has expired? I think that deserves an issue.

arbeitszeit_flask/forms.py
petsch0: done. seems nowadays, casefold() is recommended over lower() though

petsch0:
> Thanks! Looks good, I only have two suggestions.
> 
> So registration fails silently when the token has expired? I think that deserves an issue.

opened #1235

sloschert (approved):
Top!

PR #1204, opened 2025-03-28
Several changes to dev docs and README

sloschert:
- Move user roles from README to "concepts" in dev docs
- Add table of allowed transfers to "concepts"
- Delete "datamodel" section, because much of it is explained in "concepts"
- Move info config of webserver to new section "hosting"
- New ordering of menu items in `index.rst`


docs/concepts.rst
```diff
@@ -135,4 +153,73 @@ While this implementation may seem undemocratic at first glance, it must be note
...
+     - r
+     - On approval of a public plan, the planned hours for liquid means of production are subtracted from the PSF account and added to the R account of the company.
+   * - credit_public_a
+     - psf
+     - a
+     - On approval of a public plan, the planned hours for labor are subtracted from the PSF account and added to the A account of the company. 
+   * - private_consumption
+     - member
+     - prd
+     - On private consumption, the "coop price" is subtracted from the member's account and added to the PRD account of the producing company.
```
petsch0, line 201: why is "**coop**" in "coop price" always used here and in the following? It slightly confused me when reading.

docs/concepts.rst
```diff
@@ -135,4 +153,73 @@ While this implementation may seem undemocratic at first glance, it must be note
...
+   * - productive_consumption_p
+     - p
+     - prd
+     - On productive consumption of fixed means of production, the "coop price" is subtracted from the P account of the consuming company and added to the PRD account of the producing company.
+   * - productive_consumption_r
+     - r
+     - prd
+     - On private consumption of liquid means of production, the "coop price" is subtracted from the R account of the consuming company and added to the PRD account of the producing company.
```
petsch0, line 209: should be "On **productive** consumption" instead of "On private consumption" right?

petsch0 (commented):
this is very helpful, especially the new "transfer" table. great work @sloschert! Just two minor comments.

docs/concepts.rst
sloschert: yes, thanks for catching.

docs/concepts.rst
sloschert: yeah, i understand. it is meant to refer to the "cooperative price" explained above in this document. and technically it's accurate, because the cooperative price is the average cost, which is the amount subtracted from the member's account. it is an important information, because it it is not the "individual price" that gets subtracted. but i understand it is confusing here. have a look at the change - does that make things clearer?

sloschert: @petsch0 ping

petsch0: > does that make things clearer?

yes, it does. I like the usage of "cost" compared to "price". It kind of forces you to understand where the amount you are paying is coming from.

petsch0 (approved):

PR #1208, opened 2025-04-05
Use "transfers" in RegisteredHoursWorked use case

### LINKED ISSUE #1207: Update Registration of Hours Worked Use Case
- Update `RegisteredHoursWorked` (db model and business record):
  - Add new columns `transfer_of_work_certificates` and `transfer_of_taxes`
  - Delete now unncessary field `amount` (because the amount can be found via the foreign key of column transfer_of_work_certificates)
  - Create a database migration
- Update `RegisterHoursWorked` use case: Implement work certificates transfer and tax transfer
- Update three use cases that deal with showing registered hours to the user: `GetMemberAccount`, `ListRegisteredHoursWorkedUseCase`, `ShowAAccountDetailsUseCase`
- Update calculation of psf in `PublicSectorFundService`
- Update `AccountResult.joined_with_balance repository` method
- Update and write tests

sloschert:
This commit transitions the RegisteredHoursWork use case from using one "transaction" to using two "transfers". This was proposed in RFC #1175. The following changes were necessary:

- Change the db model and business record called `RegisteredHoursWorked`: The field `amount` is now unnecessary and the `transaction` field has been replaced by two `transfer` fields.
- The `RegisterHoursWorked` use case has been changed to register two `transfers` instead of one `transaction`.
- Three use cases that are dealing with showing registered hours to the users have been changed (`GetMemberAccount`, `ListRegisteredHoursWorkedUseCase`, `ShowAAccountDetailsUseCase`).
- Update calculation of psf in `PublicSectorFundService`
- Related presenters have been adapted.
- New repository methods have been added.
- A quite complex db migration has been written to add the new columns to the `RegisteredHoursWorked` table and to move relevant data from the `Transaction` table into the new `Transfer` table and reference these new data entries correctly in the `RegisteredHoursWorked` table.
- Tests have been adapted.

fixes #1207


sloschert:
Sorry for the huge Pull Request, i tried to split it up, but i did not really succeed. I'll try to improve on it next time.

PR #1220, opened 2025-04-21
Add transfer type

### LINKED ISSUE #1219: Add `transfer.type` field
It would be helpful, especially for queries, if the `transfer` records would have a field "type", that can be one of the types described in the developer documentation: https://arbeitszeitapp.readthedocs.io/en/latest/concepts.html#transfers-of-labor-time

On postgres database level, they should probably be implemented as enum types https://www.postgresql.org/docs/current/datatype-enum.html

I think that this new field is necessary, because in the future, when we will have "Undo" transfers, it is not possible to infer the transfer type from the debit and credit accounts involved.

sloschert:
- Add "type" field to transfer record 

  Transfers of labour time can only be of certain, limited types. It therefore makes sense to add an enum field "type" to the transfer record in order to distinguish between different kind of transfers without the need of elaborated database joins, etc. In the future, with arrival of undo-transfers or other new transfer types, it may be even impossible to distinguish between type of transfers without this new field.
  
  fixes #1219

- Remove old database types 

  This commit deletes the unused database types 'accounttypes' and 'useraction' from the db via a migration script.

### DIFF
--- a/arbeitszeit/records.py
+++ b/arbeitszeit/records.py
@@ -7,6 +7,8 @@ from enum import Enum
 from typing import Dict, List, Optional, Union
 from uuid import UUID
 
+from arbeitszeit.transfers.transfer_type import TransferType
+
 
 @dataclass
 class EmailAddress:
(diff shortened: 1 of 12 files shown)


petsch0 (approved):
great work!

PR #1230, opened 2025-05-18
Make transactions in db consumption tables non nullable

sloschert:
They where mistakenly nullable in the sqlalchemy table definitions. Also a migration has been added to ensure that these columns are non-nullable.


sloschert:
> LGTM! But, once again, I would like to have a better understanding of the DB schema.

Ok i think i can generate an overview in the next days. The db tool dbeaver has this functionality.

PR #1233, opened 2025-05-18
Change sphinx theme for dev docs

sloschert:
Currently the navigation on our dev docs [on readthedocs ](https://arbeitszeitapp.readthedocs.io/en/latest/introduction.html)are not very intuitive because of missing sidebar navigation.

This commit changes the theme to the sphinx-rtd-theme which should provide better integration for readthedocs.

The dev docs can be build locally via `make html && open build/html/index.html`

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -4,9 +4,9 @@ sys.path.insert(0, os.path.abspath('..'))
 
 
 project = 'arbeitszeitapp'
-copyright = '2022, Gruppe Arbeitszeitapp'
+copyright = '2025, Gruppe Arbeitszeitapp'
 author = 'Gruppe Arbeitszeitapp'
-html_theme = 'alabaster'
+html_theme = 'sphinx_rtd_theme'
 extensions = [
     'sphinx.ext.autodoc',
 ]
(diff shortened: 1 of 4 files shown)


petsch0 (approved):
lgtm, however not really sure what you mean with "missing sidebar navigation". in chrome there is this
sidebar navigation

do you mean that it is not good because it is not sticky, i.e. you loose it when you scroll down?
also iiuc, any additional nix dependency will blow up the size of the delivered software artefact right? this may (in future) also be a point to consider when adding new, not strictly required dependencies right?

sloschert:
Oh you are right, I'm just used to zoomed in to about 150% on my laptop when reading documentation, at which point the menu is simply shown at the bottom. Still I'd prefer the rtd theme, because it let's you zoom in further, it's sticky, as you say, and looks clearer IMO. The size is only 1-3 MB.
I merge and if for some reason we do not like it we can easily revert it.

petsch0:
> Oh you are right, I'm just used to zoom in to about 150% on my laptop when reading documentation, at which point the menu is simply shown at the bottom. Still I'd prefer the rtd theme, because it let's you zoom in further, it's sticky, as you say, and looks clearer IMO. The size is only 1-3 MB. I merge and if for some reason we do not like it we can easily revert it.

it looks indeed much better now, thanks sebas!

Issue #1229, opened 2025-05-11
Update Private Consumption Use Case


sloschert:
For context see https://github.com/ida-arbeitszeit/arbeitszeitapp/issues/1175
 
- Update any records specific to this use case.
- Implement transfers from Member to PRD
- Implement transfers between PRD and Cooperation accounts (compensations)
- Update tests to reflect the new transfer model

PR #1234, opened 2025-05-21
Show compensations in ShowPRDAccountDetailsUseCase

sloschert:
**Show compensations in ShowPRDAccountDetailsUseCase** 

In order to be able to test this implementation we must create transfers directly instead of making use of the PrivateConsumption use case, which would be preferable, because compensation transfers are not yet implemented in that use case.

Part of https://github.com/ida-arbeitszeit/arbeitszeitapp/issues/1229

### DIFF
--- a/arbeitszeit/use_cases/show_prd_account_details.py
+++ b/arbeitszeit/use_cases/show_prd_account_details.py
@@ -6,7 +6,14 @@ from decimal import Decimal
 from itertools import accumulate
 from uuid import UUID
 
-from arbeitszeit.records import AccountOwner, AccountTypes, Company, Member, Transaction
+from arbeitszeit.records import (
+    AccountOwner,
+    AccountTypes,
+    Company,
+    Cooperation,
+    Member,
+    Transaction,
+)
 from arbeitszeit.repositories import DatabaseGateway
 from arbeitszeit.transfers.transfer_type import TransferType
 
@@ -26,11 +33,17 @@ class CompanyPeer:
 
 
 @dataclass
+class CooperationPeer:
+    name: str
+    id: UUID
+
+
+@dataclass
 class TransferInfo:
     type: TransferType
     date: datetime
     volume: Decimal
-    peer: MemberPeer | CompanyPeer | None
+    peer: MemberPeer | CompanyPeer | CooperationPeer | None
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


arbeitszeit/use_cases/show_prd_account_details.py
```diff
@@ -74,20 +96,59 @@ def show_details(self, request: Request) -> Response:
...
-    def _get_credit_transfers(self, company: Company) -> list[TransferInfo]:
-        credit_transfers = self.database.get_transfers().where_account_is_debtor(
-            company.product_account
+    def _get_credit_and_compensation_for_coop_transfers(
+        self, company: Company
+    ) -> list[TransferInfo]:
+        debit_transfers_and_creditor = (
+            self.database.get_transfers()
+            .where_account_is_debtor(company.product_account)
+            .joined_with_creditor()
```
Amittai-Aviram, line 105: Does this query follow a pattern that we are likely to have to re-use for many different transfers?  If so, then perhaps we should try later to pull it out into a general utility.  Yes, it's only one line of code, but it implies a sequence of operations.

arbeitszeit/use_cases/show_prd_account_details.py
```diff
@@ -74,20 +96,59 @@ def show_details(self, request: Request) -> Response:
...
+                )
+            elif transfer.type == TransferType.compensation_for_coop:
+                transfers.append(
+                    TransferInfo(
+                        type=TransferType.compensation_for_coop,
+                        date=transfer.date,
+                        volume=-transfer.value,  # negative because account is debit account
+                        peer=self._create_peer_info(creditor),
+                    )
+                )
```
Amittai-Aviram, line 130: I wonder whether this logic can be expressed with two list comprehensions that we would then concatenate, rather than a complex for-loop — and whether that would improve readability.  (I don't know.)

Amittai-Aviram (approved):
LGTM!
I just noted a couple of small thoughts.

arbeitszeit/use_cases/show_prd_account_details.py
sloschert: For now I don't see potential for it, but we surely will reevaluate when all transactions have been "changed" to transfers.

sloschert:
Thanks for your review, Amittai!

PR #1237, opened 2025-05-29
Rely on use cases in data generator

sloschert:
Before this change the test data generator "CooperationGenerator" would rely on database operations to assign plans to a cooperation. Instead, it should rely on the use case path, because it prevents data generator user from unsupported operations like adding public plans to cooperations, adding plans to several cooperations, etc. The Generator has been changed. Moreover, the PlanGenerator has been refactored.

### DIFF
--- a/tests/data_generators.py
+++ b/tests/data_generators.py
@@ -183,6 +183,9 @@ class PlanGenerator:
         hidden_by_user: bool = False,
         rejected: bool = False,
     ) -> UUID:
+        assert not (
+            requested_cooperation and cooperation
+        ), "You cannot request a new cooperation for a plan that already belongs to one."
         if planner is None:
             planner = self.company_generator.create_company()
         draft = self.draft_plan(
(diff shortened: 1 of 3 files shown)


tests/data_generators.py
```diff
@@ -183,6 +183,9 @@ def create_plan(
         hidden_by_user: bool = False,
         rejected: bool = False,
     ) -> UUID:
+        assert not (
+            requested_cooperation and cooperation
+        ), "You cannot request a cooperation being part of it."
```
Amittai-Aviram, line 188: Not quite clear.  If the caller passes in both a UUID for a requested cooperation and one for a cooperation, does that mean that it is attempting to create a plan that belongs to a cooperation but associate it with another (or the same) cooperation?  I could suggest a clearer syntax, but I'd have to understand the error scenario better.

sloschert (commented):
Please see my answer to your comment. There was no issue for this change, but I think it's important to have a solid interface for data generators, which we use a lot in our tests. I thought the error message when providing two contradicting arguments would prevent user errors.  It is also important that data generators are using use cases for their inner logic. On top of  the data generators file is actually a comment: 

"""Data generators defined in this module *should* only interact with
use case objects.  Some of the classes in this module violate this
rule though. Feel free to change them so that they comply.
"""

tests/data_generators.py
sloschert: You are right, it is poorly phrased. The point is: When a plan is in a cooperation, it cannot request another cooperation. So, as a user of the PlanGenerator, you should be able to call the "create_plan" method with the "requested_cooperation" argument, which will create a plan that requests a cooperation (but the coordinator has not yet answered the request) OR to call it with the "cooperation" argument (which means that the created plan will be in a cooperation), but you shouldn't be able to provide both arguments.

More accurate would be maybe "A plan cannot be part of a cooperation and requesting one." Or what do you think?

The PlanGenerator, like all other data generators, are currently used mainly in tests, but also in arbeitszeit_benchmark (benchmark messurements) and in the future maybe in commands that can easily populate the dev database with test data for manual (db migration) tests, presentations, etc.

Amittai-Aviram: "You cannot request a new cooperation [planning association] for a plan that already belongs to one."

Amittai-Aviram (approved):
With the revised error message (please see my response), LGTM.

PR #1240, opened 2025-06-07
Allow unlimited overdraw of member accounts

sloschert:
After this change, by setting the variable ALLOWED_OVERDRAW_MEMBER to `-1`, the allowed overdraw of member accounts can be configured to be unlimited.

This is useful for tests and theoretical scenarios, but also for very permissive productive scenarios.

Note that this decision practically prevents us to allow negative overdraw values other than `-1`, which the author of this commits finds acceptable, because such a feature would mean to prohibit the spending of existing labour time tokens.

### DIFF
--- a/arbeitszeit/control_thresholds.py
+++ b/arbeitszeit/control_thresholds.py
@@ -2,6 +2,8 @@ from typing import Protocol
 
 
 class ControlThresholds(Protocol):
-    def get_allowed_overdraw_of_member_account(self) -> int: ...
+    def get_allowed_overdraw_of_member_account(self) -> int | None:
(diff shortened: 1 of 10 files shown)


Amittai-Aviram:
Top-level comment:  Would you consider defining a symbolic constant, INFINITE_OVERDRAW = -1 ?

sloschert:
> Top-level comment: Would you consider defining a symbolic constant, INFINITE_OVERDRAW = -1 ?

@Amittai-Aviram I agree that "-1" was not very intuitive. Please have a look at the new solution I just force pushed. It uses `None` in our python code and the literal string "unlimited" on the server. I explained it in the commit message.

PR #1310, opened 2025-10-18
Compatibility with SQLite

### LINKED ISSUE #1306: Make app compatible with postgres and sqlite
Test, dev and prod setup would be easier, if we would support not only postgres but also sqlite databases. 
There is not much postgres specific code, so it should be doable.

sloschert:
This PR makes arbeitszeitapp compatible with SQLite in development, testing and production.

A few things to consider when reviewing this PR:
- **Dev Setup**: Becomes a lot easier and faster because the usually somewhat tricky local PostgreSQL configuration is not necessarily required anymore. A "Quickstart" guide for new developers has been added with simplified instructions, which should not take longer than a few minutes to follow.
- **Developing**: Could get more complicated because we have to consider two database backends now, of which one (SQLite) is less versatile. 
- **Testing**: 
   - When running tests on both databases, test time increases obviously considerably. ("nix flake check": 4min → 10min)
   - Developers might forget to run tests with both databases. This is not critical because "nix flake check" runs both databases as part of our Github CI tests, but it makes the development/testing process less straightforward. 
   - Local tests on SQLite need only 50% of the time to run compared to Postgres
- **Production**: I'm not really sure if an Arbeitszeitapp SQLite production instance makes sense. Probably SQLite works fine for small to middle large communities, but PostgreSQL scales better for larger communities. We could compromise and support SQLite only for dev and testing while restricting production to PostgreSQL, but then the overhead of supporting two backends is perhaps not worth it.

In conclusion, I am not 100% sure if we should commit to SQLite - what are your thoughts?

fixes #1306

---

**Benchmarks** (average_execution_time_in_secs)

This branch, SQLITE:

show_prd_account_details: 0.5224635380000109
show_r_account_details: 0.07902885919993423
get_statistics: 0.6570295559999977
get_company_summary: 0.6164003847999993
query_plans_sorted_by_activation_date: 5.907044

This branch, POSTRGRES:

show_prd_account_details: 0.201217610599997
show_r_account_details: 0.09606246140001531
get_statistics: 2.293858903799992
get_company_summary: 1.496143248000044
query_plans_sorted_by_activation_date: 9.217337167000005

### DIFF
--- a/docs/development_setup.rst
+++ b/docs/development_setup.rst
@@ -1,7 +1,31 @@
 Development Setup
 =================
 
-The recommended development environment is Linux. 
+Quickstart
+-----------
+
+The following steps should get you quickly set up. You find more detailed instructions in the rest of this
+document.
+
+- The recommended development environment is Linux.
+- Clone the repository from Github.
+- Activate a virtual environment and run ``pip install -r requirements-dev.txt`` to install the dependencies.
+- Set the following environment variables in the terminal:
+
+.. code-block:: bash
+
+  export FLASK_APP=arbeitszeit_development.development_server:main
+  export ARBEITSZEITAPP_SERVER_NAME=127.0.0.1:5000
+  export ARBEITSZEITAPP_CONFIGURATION_PATH=${PWD}/arbeitszeit_development/development_settings.py
+  export DEV_SECRET_KEY="my_secret_key"
+  export ALEMBIC_CONFIG=${PWD}/arbeitszeit_development/alembic.ini
+  export ARBEITSZEITAPP_TEST_DB=sqlite:///${PWD}/arbeitszeitapp_test.db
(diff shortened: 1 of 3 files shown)


Amittai-Aviram:
This is a difficult decision!  When I first saw you mention this idea (as an issue), I thought it was great!  SQLite is widely used, and small-ish worker co-ops are perhaps more likely to be interested in switching to labor-time account for exchanges with other worker co-ops — and they are perhaps likely to use SQLite.  But you wrote the issue on the assumption that the changes would be few and fairly easy, and I think what you have found by experience is that the addition of support for SQLite costs a lot more than expected, especially in testing time, but also in the complexity of development.

I think I am inclined to say yes, let's go for it, now that you have made all these changes, and we can see how burdensome the additional testing time and development complexity becomes.  I still expect the latter to be not that bad (but I have not yet looked at your code changes).

Issue #1249, opened 2025-06-13
delivery strategy for arbeitszeitapp


petsch0:
## why?
currently, we are completely relying on [nix](https://nixos.org/) for setup of development environment and delivery/deployment of our app. this is great if you are familiar with nix, but since Jordi left we are kind of lacking that knowledge (at least for the moment). Moreover, a [nix flake](https://nixos.wiki/wiki/Flakes) is currently the only software artefact we are delivering which means that for anybody not familiar with nix, deployment of our app will be challenging. 
## current delivery process (as far as I understand it)
- in [arbeitszeitapp repo](https://github.com/ida-arbeitszeit/arbeitszeitapp) there is [flake.nix](https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/master/flake.nix) exposing the arbeitszeitapp python package via [overlays.default](https://github.com/ida-arbeitszeit/arbeitszeitapp/blob/81e78defc75b699a5b26c9a914a1fe83e42eb112/flake.nix#L69)
- in [arbeitszeitapp-deployment repo](https://github.com/ida-arbeitszeit/arbeitszeitapp-deployment) another [flake.nix](https://github.com/ida-arbeitszeit/arbeitszeitapp-deployment/blob/development/flake.nix) implements a nix module. It pulls in [arbeitszeitapp.overlays.default](https://github.com/ida-arbeitszeit/arbeitszeitapp-deployment/blob/de1af32fa490f7f0402982af6303d4705ad89db3/flake.nix#L133) and passes it to [default.nix](https://github.com/ida-arbeitszeit/arbeitszeitapp-deployment/blob/development/modules/default.nix) which in turn adds it to the system's nixpkgs overlays. Additionally, default.nix provides options for arbeitszeitapp deplyoment, as well as managing commands, config files with default values, and other required services and their configs for deploying the arbeitszeitapp, i.e. uwsgi, nginx, and postgresql. 
- this module is then finally imported in another flake.nix which sets the required config values for the arbeitszeitapp and which is used by [nixops](https://nixos.wiki/wiki/NixOps) to deploy everything.

**In summary, this is a great, well-tested, end-to-end deployment process for us, but for others it may not be so easy to reproduce since they would have to set up their own nix program consuming our flake (like in arbeitszeitapp-deployment), or use our arbeitszeitapp-deployment repo which however kind of enforces to be consumed by a system running nix os, and probably involves our assistance.** 
## alternatives
[...]

PR #1327, opened 2025-11-06
create initial Dockerfile

petsch0:
#1249 
recommendation is to have Docker Desktop installed as prerequisite.

using this Dockerile, a docker image from arbeitszeitapp can be built via
`docker build -t arbeitszeitapp:latest .` (executed from repo root)

afterwards, you will see the "arbeitszeitapp" docker image in Docker Desktop under "Images" tab (currently ~1GB which should be reduced)

to run the image in a container, this command worked for me:
`docker run -d -p 127.0.0.1:5000:5000 -e ARBEITSZEITAPP_CONFIGURATION_PATH=/arbeitszeitapp/arbeitszeit_development/development_settings.py -e ARBEITSZEITAPP_DEV_DB="postgresql://postgres@host.docker.internal:5432/Arbeitszeitapp_dev" -e ARBEITSZEITAPP_SERVER_NAME="127.0.0.1:5000" -e ALEMBIC_CONFIG=/arbeitszeitapp/arbeitszeit_development/alembic.ini -e ALEMBIC_SQLALCHEMY_DATABASE_URI="postgresql://postgres@host.docker.internal:5432/Arbeitszeitapp_dev" -e FLASK_DEBUG=1 --name arbeitszeitapp arbeitszeitapp:latest`

but yours can be different, depending on your db setup (I have the postgres dbs running on port 5432). The "-e" parameters are to inject env variables in the container at runtime. You will see the container running in Docker Desktop under "Containers" tab, and the app can now be accessed under localhost:5000.

Note that I am not sure at all anymore if this Dockerfile is helpful at the moment. This is because the Dockerfile itself should be production-ready, but in the "docker run" command from above it is used for development environment. But for dev environment we already have nix which is working just fine (so no Dockerfile required for dev). For deploying the app via docker in production on the other hand, the Dockerfile is not enough. You would need a "docker compose" setup similar to the one in
https://github.com/ida-arbeitszeit/arbeitszeitapp-deployment/pull/140
which includes tests, and reflects the nix deployment pipeline.


sloschert:
Probably that makes sense, even though it hurts, because quite some work went into it from your part. But yes, we can come back to it, whenever we decide to offer a docker-based deployment option.

